PR #62, opened 2015-08-28
wip: WS Addressing minimalist support 

Knarf64:
TODO: 

- finish doc
- support EndpointReference type instead of using a <address> for this kind of data
- support relationship and referenceParameters elements

Later on : 
- WSDL generation to follow spec, using this API


Knarf64:
I've corrected it, I'll add some specification support and propose it again.

PR #66, opened 2015-09-14
Fix #65 (error because of cyclic imports)

### LINKED ISSUE #65: error in installation : ImportError: cannot import name ForwardHeaderGenerator
I'm running python 2.7.5 on Ubuntu 14.04
I'm trying to install the library KD Soap  and I get this error : 

user@server:/path/KDSoap-master/autogen$ python autogen.py 
Traceback (most recent call last):
  File "autogen.py", line 11, in <module>
    from header import ForwardHeaderGenerator
  File "/path/KDSoap-master/autogen/header.py", line 6, in <module>
    import autogen
  File "/path/KDSoap-master/autogen/autogen.py", line 11, in <module>
    from header import ForwardHeaderGenerator
ImportError: cannot import name ForwardHeaderGenerator

ByteCommander:
This should fix the error (issue #65):

    $ python autogen.py 
    Traceback (most recent call last):
      File "autogen.py", line 11, in <module>
        from header import ForwardHeaderGenerator
      File "/path/KDSoap-master/autogen/header.py", line 6, in <module>
        import autogen
      File "/path/KDSoap-master/autogen/autogen.py", line 11, in <module>
        from header import ForwardHeaderGenerator
    ImportError: cannot import name ForwardHeaderGenerator

Please verify this edit and apply it to your repository if it solves that problem.

### DIFF
--- a/autogen/autogen.py
+++ b/autogen/autogen.py
@@ -8,7 +8,7 @@ if sys.version_info[0] != 2:
 
 from cpack import CPackGenerateConfiguration 
 from configure import ConfigureScriptGenerator
-from header import ForwardHeaderGenerator
+import header
 
 def parseSvnInfo( stdout ):
 	lines = stdout.splitlines()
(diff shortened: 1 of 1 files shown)


ByteCommander:
The author of issue #65 who also posted this problem as question on AskUbuntu.com confirmed that my edit fixes the problem. See my [answer](http://askubuntu.com/a/674012/367990) there.

winterz:
yeah but why are you doing that?  you should be running python autogen.py from the top-level source or from a build dir

winterz:
seems a better idea is to rename the autogen/autogen.py module so folks don't confuse it with the top-l
level autogen.py program

ByteCommander:
@winterz Yes, after taking a look again and finding another `autogen.py` in the repository's root directory, I agree with you that you should rather rename `/autogen/autogen.py` and make it clear that this file is *not* supposed to be run than applying my edit. Maybe you even want to add a line that prints an error message if somebody tries to run it?

    if __name__ == "__main__":
      print "Error! You are not supposed to run `.../autogen/autogen.py`."
      print "Please use `autogen.py` directly from the project's root directory instead."

PR #68, opened 2015-10-01
WS Addressing Support

Knarf64:
The ws addressing feature is about setting endpoint and other metadata
into the soap header.

We setup a test wsaddressingtest to cover the direct KDSoapClientInterface API
and will extend it to WSDL generation when needed.
KDSoapClient has a new  abstraction class for ws addressing : KDSoapMessageAddressingProperties
which is a simple layer carrying header soap information. KDSoap message is holding a KDSoapMessageAddressingProperties
that is used when writing the soap envelope Futur patch will use this api to generate addressing properties from WSDL file.
Adding KDSoapEndpointReference to KDSoapClient

References :
http://www.w3.org/TR/ws-addr-core/
http://www.w3.org/TR/ws-addr-soap/
http://www.w3.org/TR/ws-addr-wsdl/

### DIFF
--- a/doc/CHANGES_1_5.txt
+++ b/doc/CHANGES_1_5.txt
@@ -4,7 +4,7 @@ General:
 
 Client-side:
 ============
-*
+* Add WS Addressing Support
 
 Server-side:
 ============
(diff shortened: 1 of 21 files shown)


Knarf64:
I updated it with corrections,  working to finish the tiny missing spec part.

Knarf64:
updated

Knarf64:
updated

PR #70, opened 2015-10-03
Make the SSL error test pass with yet another signal variation.

hporten:
Getting just these two with Qt 5.4.2 on Debian 'stretch'. Same
as the comment suggests for Mac OS X. Question is which system/Qt
combination had 3 errors.

(This is a rather lame solution and more of a bug report than a bugfix. If the results are really that unportable the check might be easier to dumb down the check and make it just check presence
of 1-2 mandatory errors)


hporten:
Fine then. And yes, it's me :) I'll explain why I ran the tests next week.

PR #74, opened 2016-01-05
Kdwsdl2cpp path

phbaer:
This fixes a compile error (wsdl stub implementation not generated) with MSVC (nmake, msbuild/devenv)


dfaure-kdab:
Strange that our CI didn't catch that... let's see what happens when this get in ;)

phbaer:
There was no immediate error message shown but when MOC'ing the generated header, moc.exe returned an error (Win7, nmake or msbuild of MSVC 2012/2015). It was quite a strange behaviour... the first fix I committed was only suitable for Qt5 ($$shell_path) and hence I did not want to create a pull request for that.

Let's see what the next CI tells us. Thanks for merging and thanks for KDSoap, really appreciated! :)

PR #75, opened 2016-02-07
Added postfix d to win32 debug libraries

githubuser0xFFFF:


winterz:
fixed in e4da45bce074a0249f2b636bff2797a2404d1a95.
I used a slightly different patch than yours

Issue #82, opened 2016-05-16
kdwsdl2cpp: Type/name collisions in nested complexTypes


hkaelber:
We encountered some problems using kdsoap for accessing ciscos axl soap interface [1] via kdwsdl2cpp generated classes. The used schema [2] defines tons of nested complex-types within elements having the same name, while the structure of the complexTypes are different.

Take as an example the <xsd:element name='return'> that is used in many places as container for API-responses:
cat AXLSoap.xsd | egrep "name=[\'\"]return[\'\"]" | wc -l 
292

Using kdwsdl2cpp as is, this leads to a single generated type XSD__Return defined following the first occurence of <xsd:element name='return'> within the xsd, which is nested in <xsd:complexType name='GetSipProfileRes'>. Now if you want to evaluate other responses that use return you will only be able to access a XSD__Return with the wrong members (those of 'GetSipProfileRes').

The quick-fix in commit [3] resolves the issue for me. It simply checks whether a complexType of a given name already exists and increments a numeric postfix until we have a new type-name.

This is not an optimal solution because thus also different classes are generated for complexTypes that are structurally identic, which is not wrong, but further bloats the generated code, in my case:

Before:
Converting 141 simple types 
Converting 1563 complex types
18M  wsdl_AXLAPI.cpp
3.0M wsdl_AXLAPI.h

After:
Converting 145 simple types 
Converting 2420 complex types
21M  wsdl_AXLAPI.cpp-new
3.6M wsdl_AXLAPI.h-new

A better solution could check for complete structural identity (via a ComplexType::operator==() or so) instead of typename equality.

What do you think?

Thanks,
  Holger

[1] https://developer.cisco.com/site/axl/
[2] https://developer.cisco.com/media/axl-schema-11-0/AXLSoap.xsd
[3] https://github.com/hkaelber/KDSoap/commit/9913e80031dc80009aaaf922ddd03b091438d5a8

dfaure:
Hello, sorry for the delay in my reply.

I agree with your analysis and solutions (the quick one in the mentioned commit and the idea for a better one).

What's missing is a unittest triggering this case. I tested your patch, and none of the WSDL/XSD files under unittests led to the qDebug() being shown. Can you either modify the WSDL in e.g. unittest/wsdl_document so that it hits this issue, or send me the full XSD file, assuming I can make it public (by adding it to a kdsoap unittest) ?

hkaelber:
Pull request for simple fix:
[...]

PR #91, opened 2016-10-18
kdwsdl2cpp: Avoid potential type collisions in nested complexTypes

hkaelber:
So far, for complexTypes nested in elements with the same name a
single C++ class has been created. This is wrong in cases where
multiple elements with the same name contain complexTypes of different
structure because different classes should be created

https://github.com/KDAB/KDSoap/issues/82

### DIFF
--- a/kdwsdl2cpp/schema/parser.cpp
+++ b/kdwsdl2cpp/schema/parser.cpp
@@ -520,8 +520,13 @@ Element Parser::parseElement(ParserContext *context,
             //qDebug() << "childName:" << childName.localName();
             if (childName.localName() == QLatin1String("complexType")) {
                 ComplexType ct = parseComplexType(context, childElement);
-
-                ct.setName(newElement.name());
+                QString name = newElement.name();
+                int i = 0;
+                while (!d->mComplexTypes.complexType(QName(ct.nameSpace(), name)).name().isEmpty())
+                    name = newElement.name() + QString::number(++i);
+                if (name != newElement.name() && debugParsing())
+                    qDebug() << " detected potential type collision in nested complexType, updated name" << newElement.name() << "to" << name;
+                ct.setName(name);
                 ct.setAnonymous(true);
                 d->mComplexTypes.append(ct);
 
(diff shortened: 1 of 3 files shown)


hkaelber:
Thanks for the comment, I agree.

I pushed a commit, that implements some operator==().

I updated the test-cases a bit and made sure that your example from logbookifv3.wsdl is covered and did some tests with our cisco AXL wsdls.

I decided to define equality with respect to the corresponding data-structures/the XML schema definitions and not the resulting generated classes. That means that there can be cases of nested complex types that are considered inequal due to some properties (like i.e. Element.min/maxOccurs()) that lead to the same generated classes.

After some testing with the gigantic cisco AXL API, this seems to be ok, because most such cases that I saw seemed to have little impact with respect to generated code size.

If you want to further optimize this, pls point me to the properties that I could probably drop from the operator==().

Nonetheless, as this further bloats the generated code (in case of the AXL API v. 11.0 to > 25MB, which does not compile on my 4 gig RAM machine without swapping :-( I'd love to add some --whitelisted-operation="op1,op2,..." parameter to kdwsdl2cpp to add a means to limit the size of generated code by telling it which operations should be generated. Some quick tests looked promising. Comments from your side?

Thanks,
  Holger

dfaure-kdab:
Wow I seem to have completely missed that last comment, in almost two years... Email overload.

I like the idea of specifying which operations to generate, for the case of a huge wsdl file where only some operations are wanted. Maybe instead of a (huge) command-line argument it could be a sort of "config file" (better suited for the case of a very long list of operations). This would also open up the possibility of manually specifying namespace prefixes in that same config file, which is something others have asked about as well. This is something that can of course be implemented completely separately from what is being discussed here.

But I wonder if by now you're still interested... really sorry about the huge delay in my reply.

raphaelcotty:
I would be interested by the patch.
I see that most of the changed now belong to the https://github.com/cornelius/libkode repository.
Would that make sense if I try to reactivate the patch?

dfaure-kdab:
Yes it would certainly make sense to rebase/redo the patch on top of the libkode repository.

PR #92, opened 2016-11-23
When an optional list is deserialized, the associated nil variable is…

diman1234:
… set to false

### DIFF
--- a/unittests/optionaltype_pointer/testpointerapi.cpp
+++ b/unittests/optionaltype_pointer/testpointerapi.cpp
@@ -25,6 +25,7 @@
 #include "testpointerapi.h"
 #include <QtTest/QtTest>
 #include <QDebug>
+#include <QList>
 
 TestPointerApi::TestPointerApi()
 {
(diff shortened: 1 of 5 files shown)


diman1234:
I've added the unittest.

unittests/optionaltype_pointer/testpointerapi.cpp
```diff
@@ -39,4 +40,20 @@ void TestPointerApi::test()
...
+    TNS__TestOperationResponse1 resp;
+    QCOMPARE(resp.outOptionalArray(), (QList<qint64> *)0);
+    const QList<qint64> valIn(QList<qint64>() << 1 << 2 << 3);
+    resp.setOutOptionalArray(valIn);
+
+    TNS__TestOperationResponse1 resp2;
+    resp2.deserialize(resp.serialize("array"));
+
+    QVERIFY(resp2.outOptionalArray());
+    if (resp2.outOptionalArray()) {
```
dfaure-kdab: This if() isn't needed, since QVERIFY returns on error. So if we're here, the condition in this if() can't ever be false.

unittests/optionaltype_pointer/testpointerapi.cpp
```diff
@@ -39,4 +40,20 @@ void TestPointerApi::test()
     QCOMPARE(*resp.out(), newVal);
 }
 
+void TestPointerApi::testOptionalArray()
+{
+    TNS__TestOperationResponse1 resp;
+    QCOMPARE(resp.outOptionalArray(), (QList<qint64> *)0);
```
dfaure-kdab: QVERIFY(!resp.outOptionalArray()) would be enough (and would avoid the C cast, to which I'm allergic due to bad experiences ;) )

diman1234:
:-) done

PR #97, opened 2017-01-03
Ensure the SslHandler has been created before deleting it

sandvige:


dfaure-kdab:
Can you explain in which case this was a problem?

The commit log seems to indicate a misunderstanding, because "delete NULL" is a perfectly valid operation in C++ (which does nothing).

So the commit log that describes this change would rather be "Use deleteLater instead of delete for the SSL handler", but why? Is there a case where it is still in use after this call?

sandvige:
Hello,

You're right, the commit message is a bit clumsy :)

Since calls to the SslHandler can be delayed (see the comments for KDSoapClientInterface::sslHandler() method), it is possible to miss some of them if the client is deleted (which can be the case since it is not a QObject and cannot be deleted using deleteLater())

dfaure-kdab:
I'm still confused about how this can happen. Do you have a testcase or a backtrace or more details?

sandvige:
When calls to the SSL Handlers are queued, it can be a problem to roughly delete the receiver instead of using deleteLater(), because it will ignore/discard pending calls

sandvige:
@dfaure-kdab ping?

dfaure-kdab:
Sorry this is still all very unclear to me.

First, the commit log needs to be amended, since "has been created" is definitely not the issue, as we've established.

Then, you last comment says deleting will ignore/discard pending calls. Of course it does, but why is that a problem? The whole client is going away anyway, why do we want to still process signals from or to the SSL handler? Please give more information about what you're doing, what is happening and why.
I don't mean to be difficult, I'm just trying 1) to apply the general rule of always understanding a commit, 2) to make sure that this is indeed a problem in KDSoap and not in the way it's being used...

dfaure-kdab:
Do you have a testcase for the issue?
Or failing that, a valgrind log showing the issue?

dfaure-kdab:
@sandvige ping?

sandvige:
@dfaure-kdab You're totally right, it is very important to catch every piece of a commit. Sorry, we don't have that much time as of today, and this issue is quite old for now. If I remember correctly, we had the widget displaying the associated SSL error into another thread, and it was randomly crashing. This is only what I remember.

Issue #106, opened 2017-07-11
Unexpected character


Massimo-B:
We found an issue with a WSDL file that is referencing remote XSD code. While running in a proxied environment we use tsocks to connect:

```
tsocks ~/bin/kdwsdl2cpp -o Extensions.h Extensions.wsdl 
Warning: Definitions: unknown tag wsp:UsingPolicy
Warning: Definitions: unknown tag wsp:Policy
Warning: Definitions: unknown tag wsp:Policy
importing schema at http://www.unece.org/cefact/namespaces/StandardBusinessDocumentHeader
Downloading 'http://www.unece.org/cefact/namespaces/StandardBusinessDocumentHeader'
13:33:05 libtsocks(12842): Call to connect received on completed request 10
Download successful
Error[1:50] unexpected character
Warning: PortType: unknown tag wsp:Policy
Warning: Operation: unknown tag wsp:Policy
No service tag found in the wsdl file, generating one service per binding 
Converting 0 simple types 
Converting 0 complex types 
```

There are 2 issues, that do not seem to be proxy related:

**Error[1:50] unexpected character**
Is this error in the XSD or in the WSDL file? I can't see something in line 1, column 50.

**No service tag found in the wsdl file**
Do you need the WSDL for further investigation? There is actually no <service> in the WSDL. I guess this is just a WSDL derived from a plain XSD as kdwsdl was not able to import XSD directly. Is a <service> required in general just for defining data structures?

Kind regards,
Massimo

dfaure-kdab:
I strongly suggest that you make a one-time download of the XSD file(s) and adjust your WSDL to point to the file(s) locally, as I did in all unittests. This saves bandwidth usage and would also fix your problems with tsocks ;-)

The parse error "unexpected character" sounds like it's from the downloaded XSD, but I'd say, this isn't worth debugging, just have a local copy of the file.

"No service tag" is fine, as the line says, we're generating one automatically, so that shouldn't be a problem.

What might be a problem though, with this approach, is the code that strips out any data structures that aren't used by the WSDL. See Converter::cleanupUnusedTypes.
If you add a command-line option for kdwsdl2cpp to keep unused types (by skipping the call to cleanupUnusedTypes when that new option --keep-unused-types is set), I can merge that in.

PR #107, opened 2017-07-19
Add a -keep-ununsed-types argument to kdwsdl2cpp

Knarf64:
This is a follow-up to your comment from https://github.com/KDAB/KDSoap/issues/106 : 

"What might be a problem though, with this approach, is the code that strips out any data structures that aren't used by the WSDL. See Converter::cleanupUnusedTypes. If you add a command-line option for kdwsdl2cpp to keep unused types (by skipping the call to cleanupUnusedTypes when that new option --keep-unused-types is set), I can merge that in."

### DIFF
--- a/unittests/keep_unused_types/keep_unused_types.cpp
+++ b/unittests/keep_unused_types/keep_unused_types.cpp
@@ -1,0 +1,48 @@
+/****************************************************************************
+** Copyright (C) 2010-2017 Klaralvdalens Datakonsult AB, a KDAB Group company, info@kdab.com.
+** All rights reserved.
+**
+** This file is part of the KD Soap library.
+**
+** Licensees holding valid commercial KD Soap licenses may use this file in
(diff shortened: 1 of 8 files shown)


Knarf64:
updated with unit tests

unittests/keep_unused_types/keep_unused_types.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+private slots:
+
+    void testKeepUnusedTypesArgument()
+    {
+        // Unused type within WSDL, if this compiles, then -keep-unused-types works since
+        // otherwise cleanupUnusedTypes function would have removed it from the cpp generation
+        TNS__UnusedElement element;
+        Q_UNUSED(element);
+        QCOMPARE(42,42);
```
dfaure-kdab: I can't say that I find this line particularly useful ;-)

Knarf64:
updated.

PR #120, opened 2018-02-27
- Minor improvement of documentation generated by kdwsdl2cpp 

ryanenzo:


kdwsdl2cpp/src/converter_clientstub.cpp
```diff
@@ -731,14 +731,16 @@ void Converter::convertClientOutputMessage(const Operation &operation,
         const Binding &binding, KODE::Class &newClass)
 {
     // result signal
-    QString operationName = lowerlize(operation.name());
-    KODE::Function doneSignal(operationName + QLatin1String("Done"), QLatin1String("void"), KODE::Function::Signal);
-    doneSignal.setDocs(QLatin1String("This signal is emitted whenever the call to ") + operationName + QLatin1String("() succeeded."));
+    const QString operationName = lowerlize(operation.name());
+    const QString signalBase = lowerlize(operation.name());
```
dfaure-kdab: This is the same as operationName just above, why call lowerlize() twice?
Since both QStrings are exactly the same, maybe a single one is enough.
Otherwise, signalBase = operationName would at least be faster than calling lowerlize again.

ryanenzo: Two different variables for two different purposes - for easier understanding and possible future adaptions. I suppose you could use signalBase = operationName instead of calling lowerize twice, but I don't think this is something that needs to be super optimized. (Nor is that change a super saving).

dfaure-kdab: It's just good practice to not write code that is suboptimal when it's as easy to write clean and faster code. This is called "avoiding premature pessimization".

Definition: Premature pessimization is when you write code that is slower than it needs to be, usually by asking for unnecessary extra work, when equivalently complex code would be faster and should just naturally flow out of your fingers. (https://herbsutter.com/2013/05/13/gotw-2-solution-temporary-objects)

ryanenzo: And I would guess that premature optimization is what led to the badly generated documentation in the first place. I also think that clarity of code is more important over trivial optimizations like this. (Naturally, one can also argue about which is more clear, ...) But this isn't my code base. I simply provided one possible solution, it should be trivial for you to adapt the fix to what you would like to see.

dfaure-kdab: OK, I have merged your commit and then made the small change on top. Thanks for your contribution, sorry that we disagreed on a minor point, keep them coming ;-)

PR #126, opened 2018-04-13
Empty element bug fix

SudakovIvan:


unittests/empty_element_wsdl/empty_element_wsdl.cpp
```diff
@@ -0,0 +1,64 @@
...
+**
+** Contact info@kdab.com if any conditions of this licensing are not
+** clear to you.
+**
+**********************************************************************/
+
+#include "KDSoapValue.h"
+#include "wsdl_empty_element.h"
+
+#include <QtTest/QtTest>
```
dfaure-kdab: This includes the whole module *and* all of QtCore, better just include QTest (which is one single header)

SudakovIvan:
isNil() is used in  
`void KDSoapServerObjectInterface::storeFaultAttributes(KDSoapMessage &message) const`
and maybe someone will need to check the element for emptiness(has no children and attributes).
What do you think?

dfaure-kdab:
OK let's keep isNil() then. Thanks!

unittests/empty_element_wsdl/empty_element_wsdl.cpp
SudakovIvan: done

PR #127, opened 2018-04-15
don't abort the http call if authentication is not possible

caspermeijn:
If no credentials is available or if the credentials are incorrect,
then the real error message is hidden behind the cancellation of the
http call. By not aborting the call, the error state will stay
QNetworkReply::AuthenticationRequiredError. This will result is KDSoap
generating a SoapFault containing the correct error message.
This way the application can detect the error and ask the user for
correct credentials.


dfaure-kdab:
Thanks for your contribution. Any chance for a unittest for this issue?

dfaure-kdab:
If that test fails without the fix, and works with the fix, then yes, it's enough ;)

dfaure-kdab:
Pushed to master in f513e7e after improving the commit log and adding a changelog for 1.8.

Thanks for your contribution!

PR #128, opened 2018-04-29
Add support for soap 1.2 fault in faultAsString

caspermeijn:
Soap 1.2 has a different fault format then soap 1.1. This can be
detected by the namespace that the fault is in. In soap 1.2 faults are
identified by a machine-readable error code and this can be further
categorized with subcodes. It contains human-readable reason is send as
well. This change allows faultAsString to show soap 1.2 faults in a
similar way as soap 1.1 faults.

### DIFF
--- a/src/KDSoapClient/KDSoapMessage.cpp
+++ b/src/KDSoapClient/KDSoapMessage.cpp
@@ -124,12 +124,29 @@ bool KDSoapMessage::isFault() const
 
 QString KDSoapMessage::faultAsString() const
 {
-    // This better be on a single line, since it's used by server-side logging too
-    const QString actor = childValues().child(QLatin1String("faultactor")).value().toString();
-    return QObject::tr("Fault code %1: %2%3")
-           .arg(childValues().child(QLatin1String("faultcode")).value().toString())
-           .arg(childValues().child(QLatin1String("faultstring")).value().toString())
-           .arg(actor.isEmpty() ? QString() : QString::fromLatin1(" (%1)").arg(actor));
+    if (namespaceUri() == QLatin1String("http://www.w3.org/2003/05/soap-envelope"))
+    {
+        QString faultCodeStr;
+        KDSoapValue faultCode = childValues().child(QLatin1String("Code"));
+        while (!faultCode.isNull()) {
+            if (!faultCodeStr.isEmpty())
+                faultCodeStr += QLatin1String(" ");
+            faultCodeStr += faultCode.childValues().child(QLatin1String("Value")).value().toString();
+            faultCode = faultCode.childValues().child(QLatin1String("Subcode"));
+        }
+        return QObject::tr("Fault %1: %2")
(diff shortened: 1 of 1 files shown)


src/KDSoapClient/KDSoapMessage.cpp
```diff
@@ -124,12 +124,29 @@ bool KDSoapMessage::isFault() const
 
 QString KDSoapMessage::faultAsString() const
 {
-    // This better be on a single line, since it's used by server-side logging too
-    const QString actor = childValues().child(QLatin1String("faultactor")).value().toString();
-    return QObject::tr("Fault code %1: %2%3")
-           .arg(childValues().child(QLatin1String("faultcode")).value().toString())
-           .arg(childValues().child(QLatin1String("faultstring")).value().toString())
-           .arg(actor.isEmpty() ? QString() : QString::fromLatin1(" (%1)").arg(actor));
+    if(name() == QLatin1String("Fault") && namespaceUri() == QLatin1String("http://www.w3.org/2003/05/soap-envelope"))
```
dfaure-kdab: missing space after "if"

src/KDSoapClient/KDSoapMessage.cpp
```diff
@@ -124,12 +124,29 @@ bool KDSoapMessage::isFault() const
...
+    {
+        QString faultCodeStr;
+        KDSoapValue faultCode = childValues().child(QLatin1String("Code"));
+        while(!faultCode.isNull()) {
```
dfaure-kdab: This while loop confused me for a while (haha, nice pun).

I suppose Subcode elements can't be nested, this was just to factorize code between Code and Subcode handling? I think it makes the code harder to read/understand.
Can you unroll the loop, i.e. handle Code and then handle Subcode, sequentially? This will duplicate only one line of code, the child("Value") line.

And it will make it possible to format the subcode stuff differently if we want to; I'm not sure it looks great right now when having a subcode (did you test it?)

src/KDSoapClient/KDSoapMessage.cpp
```diff
@@ -124,12 +124,29 @@ bool KDSoapMessage::isFault() const
...
+            if(faultCodeStr.size())
```
dfaure-kdab: please make that

    if (!faultCodeStr.isEmpty())

src/KDSoapClient/KDSoapMessage.cpp
caspermeijn: I believe Subcode can be nested. I got that from: https://www.w3.org/TR/soap12-part1/#faultsubcodeelement

Would it help if I added a comment that explains the loop?

Yes, I did test it. For example:
> Fault s:Sender ter:NotAuthorized: Sender not Authorized. Invalid username or password!

dfaure-kdab:
Merged into commit 73d592f, sorry for the delay.

caspermeijn:
Thanks

PR #129, opened 2018-04-29
Improve detection of soap fault in http response

caspermeijn:
The soap 1.2 spec states that soap faults of type env:Sender should have
http status code 400 Bad request. The previous code assumed that all
soap error have status code 500 (which is true for soap 1.1). This
change tries to find a soap fault in the response data (regardless of
the http status code) and will return that if available. Else it will
fallback to the generic fault based on the response error.

### DIFF
--- a/src/KDSoapClient/KDSoapPendingCall.cpp
+++ b/src/KDSoapClient/KDSoapPendingCall.cpp
@@ -109,17 +109,26 @@ void KDSoapPendingCall::Private::parseReply()
 
     if (!data.isEmpty()) {
         KDSoapMessageReader reader;
-        reader.xmlToMessage(data, &replyMessage, 0, &replyHeaders);
+        reader.xmlToMessage(data, &replyMessage, 0, &replyHeaders, this->soapVersion);
     }
 
     if (reply->error()) {
-        if (!replyMessage.isFault())
-        {
+        if (!replyMessage.isFault()) {
             replyHeaders.clear();
             replyMessage = KDSoapMessage();
             replyMessage.setFault(true);
-            replyMessage.addArgument(QString::fromLatin1("faultcode"), QString::number(reply->error()));
-            replyMessage.addArgument(QString::fromLatin1("faultstring"), reply->errorString());
+            if (this->soapVersion == KDSoapClientInterface::SOAP1_2) {
+                replyMessage.setNamespaceUri(QString::fromLatin1("http://www.w3.org/2003/05/soap-envelope"));
+                KDSoapValueList codeValueList;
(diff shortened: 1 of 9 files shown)


src/KDSoapClient/KDSoapPendingCall.cpp
```diff
@@ -124,4 +111,15 @@ void KDSoapPendingCall::Private::parseReply()
         KDSoapMessageReader reader;
         reader.xmlToMessage(data, &replyMessage, 0, &replyHeaders);
     }
+
+    if (reply->error()) {
+        if(!replyMessage.isFault())
```
dfaure-kdab: missing space after if

src/KDSoapClient/KDSoapPendingCall.cpp
```diff
@@ -124,4 +111,15 @@ void KDSoapPendingCall::Private::parseReply()
...
+        {
+            replyHeaders.clear();
+            replyMessage = KDSoapMessage();
+            replyMessage.setFault(true);
+            replyMessage.addArgument(QString::fromLatin1("faultcode"), QString::number(reply->error()));
```
dfaure-kdab: That's not very SOAP-1.2, is it? ;-)
Don't you need to update this to the 1.2 elements for faults (depending on the soap version set in KDSoapClientInterface), so this works with your other commit?

caspermeijn: yes, you are right, this is not very SOAP 1.2. I was just trying to make it work in my situation :-) 
What would be the best way to make the version available in this function? I ask this because it is very easy to create a circular dependency between KDSoapPendingCall and KDSoapClientInterface.

dfaure-kdab: One solution would be to pass it as int and cast it, but that's ugly.
A better solution is to set it directly into the private class, bypassing the public API:
http://www.kdab.com/~dfaure/2017/0001-Pass-SOAP-version-to-KDSoapPendingCall.patch
Feel free to integrate this patch into yours.
Thanks!

caspermeijn: I added a second commit to support the SOAP 1.2 faults. I added this to the message reader as well. And I added unit tests for both changes and both soap versions.

dfaure-kdab: Ah, you didn't push the updated commits to this merge request, but I found the commits in your branch. They look good, I'll merge them.

Hmm, if soap-1.2 says faults don't use http status code 500, then KDSoapServerSocket.cpp needs an update as well... Are you interested in fixing the server side too?

dfaure-kdab:
All merged, thanks.

PR #132, opened 2018-06-09
additionalHttpResponseHeaderItems feature

SudakovIvan:


src/KDSoapServer/KDSoapServerObjectInterface.h
```diff
@@ -125,6 +127,17 @@ class KDSOAPSERVER_EXPORT KDSoapServerObjectInterface
      */
     virtual void processRequestWithPath(const KDSoapMessage &request, KDSoapMessage &response, const QByteArray &soapAction, const QString &path);
 
+    typedef QPair<QString/*name*/, QString/*value*/> HttpResponseHeaderItem;
```
dfaure-kdab: A struct (which means named members) is always more readable than a QPair, where one has to remember that .first means name and .second means value.

src/KDSoapServer/KDSoapServerObjectInterface.h
```diff
@@ -125,6 +127,17 @@ class KDSOAPSERVER_EXPORT KDSoapServerObjectInterface
...
+    typedef QList<HttpResponseHeaderItem> HttpResponseHeaderItems;
+
+    /**
+     * Returns additional HTTP response header items to be added to each HTTP response header
+     * The default implementation in this base class returns an empty list
+     * Subclasses can override this method as needed
+     *
+     */
+    virtual const HttpResponseHeaderItems additionalHttpResponseHeaderItems();
```
dfaure-kdab: Please remove the "const" which doesn't bring any benefit in this particular way of being used (it even prevents move semantics from applying).

src/KDSoapServer/KDSoapServerSocket.cpp
```diff
@@ -141,6 +141,20 @@ static QByteArray httpResponseHeaders(bool fault, const QByteArray &contentType,
...
+    if (serverObjectInterface) {
+        const KDSoapServerObjectInterface::HttpResponseHeaderItems& additionalItems = serverObjectInterface->additionalHttpResponseHeaderItems();
+        Q_FOREACH (const KDSoapServerObjectInterface::HttpResponseHeaderItem& headerItem, additionalItems) {
+            httpResponse += headerItem.first.toLatin1();
+            httpResponse += ": ";
+            httpResponse += headerItem.second.toLatin1();
+            httpResponse += "\r\n";
+        }
+    } else {
+        qWarning() << QString::fromLatin1("Server object %1 does not implement KDSoapServerObjectInterface!").arg(QString::fromLatin1(serverObject->metaObject()->className()))
```
dfaure-kdab: Since additional items is an optional feature, I would just remove this warning (i.e. the else branch) altogether.

src/KDSoapServer/KDSoapServerObjectInterface.h
```diff
@@ -125,6 +127,17 @@ class KDSOAPSERVER_EXPORT KDSoapServerObjectInterface
      */
     virtual void processRequestWithPath(const KDSoapMessage &request, KDSoapMessage &response, const QByteArray &soapAction, const QString &path);
 
+    typedef QPair<QString/*name*/, QString/*value*/> HttpResponseHeaderItem;
+    typedef QList<HttpResponseHeaderItem> HttpResponseHeaderItems;
```
dfaure-kdab: Please use QVector. The pair/struct is bigger than a pointer, so QList has to allocate nodes on the heap, which is slow and wasteful.

SudakovIvan: And QVector doesn't?  How does QVector allocate memory?

dfaure-kdab: QVector is like std::vector, it's a contiguous array. Especially if you resize() [or reserve()] upfront, a single allocation is needed, rather than one per 'node'.

SudakovIvan: Ok, I figured it out now, thanks

src/KDSoapServer/KDSoapServerObjectInterface.h
SudakovIvan: Well, it's a habit of the old style (c++98), thanks.
And what do you think about adding "const" at the end of the function? It seems to be better

dfaure-kdab:
OK for making the method const.
In terms of code formatting, this uses the same coding style as KDE Frameworks 5, which is the Qt coding style plus curly braces even for one-line if/for/... statements. Tool suggestions: uncrustify-kf5 (or astyle-kdelibs) in git://anongit.kde.org/kde-dev-scripts

SudakovIvan:
I'll make an unittest and tell you when I'm done

src/KDSoapServer/KDSoapServerObjectInterface.cpp
```diff
@@ -46,6 +46,14 @@ class KDSoapServerObjectInterface::Private
     QPointer<KDSoapServerSocket> m_serverSocket;
 };
 
+KDSoapServerObjectInterface::HttpResponseHeaderItem::HttpResponseHeaderItem(const QByteArray &name, const QByteArray &value): m_value(value), m_name(name)
```
dfaure-kdab: coding style: the ":" (and what follows it) should be on a separate line

unittests/servertest/servertest.cpp
```diff
@@ -1300,6 +1308,29 @@ private Q_SLOTS:
...
+        request.setHeader(QNetworkRequest::ContentTypeHeader, soapHeader.toUtf8());
+        QNetworkAccessManager accessManager;
+        QNetworkReply *reply = accessManager.post(request, rawCountryMessage());
+        QEventLoop loop;
+        connect(reply, SIGNAL(finished()), &loop, SLOT(quit()));
+        loop.exec();
+        reply->readAll();
+
+        QVERIFY(reply->rawHeaderList().contains("Access-Control-Allow-Origin"));
+        QVERIFY(reply->rawHeader("Access-Control-Allow-Origin") == "*");
```
dfaure-kdab: should use QCOMPARE (so we can see the values on failure)

unittests/servertest/servertest.cpp
```diff
@@ -1300,6 +1308,29 @@ private Q_SLOTS:
...
+        QVERIFY(reply->rawHeaderList().contains("Access-Control-Allow-Headers"));
+        QVERIFY(reply->rawHeader("Access-Control-Allow-Headers") == "Content-Type");
```
dfaure-kdab: QCOMPARE

SudakovIvan:
done

PR #133, opened 2018-06-24
kdwsdl2cpp: Added import-path argument to search for files locally

caspermeijn:
Normally files needed by kdwsdl2cpp are downloaded. This patch adds an
option to search for the files locally. This increases the compilation
speed and improved reproducibility.
When the import-path is supplied, the needed files (for example schemas)
are search at: \<importpath>/\<URL-host>/\<URL-path>
If the file is not found at the import path, then it will be downloaded
as it normally would.
For example the schema http://www.w3.org/2001/xml.xsd with the option
"-import-path /tmp" will be search at /tmp/www.w3.org/2001/xml.xsd.
The benefit is that the needed schemas can be pre-downloaded to increase
compilation speed or can be supplied with the wsdl to increase
reproducibility.


kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -63,6 +65,17 @@ bool FileProvider::get( const QUrl &url, QString &target )
       return true;
   }
 
+  QString importPath = Settings::self()->importPath();
+  if (!importPath.isEmpty()) {
+      QDir importDir(importPath);
+      QString path = importDir.absoluteFilePath(url.host() + QDir::separator() + url.path());
+      if(QFile::exists(path)) {
```
dfaure-kdab: Missing space after if

kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -63,6 +65,17 @@ bool FileProvider::get( const QUrl &url, QString &target )
...
+          qDebug("Using import path '%s'", qPrintable(path));
```
dfaure-kdab: I think this should be removed, or made more complete by printing url as well, to explain which file was found locally. If there are many files to look up, this output isn't very useful.

kdwsdl2cpp/src/main.cpp
```diff
@@ -125,6 +129,13 @@ int main(int argc, char **argv)
             }
         } else if (opt == QLatin1String("-keep-unused-types")) {
             keepUnusedTypes = true;
+        } else if (opt == QLatin1String("-import-path")) {
+            ++arg;
+            if (!argv[arg]) {
+                showHelp(argv[0]);
+                return 1;
+            }
+            importPath = argv[arg];
```
dfaure-kdab: Shouldn't this be a list, so people want to specify -import-path multiple times, in case they have multiple local paths to search into? Alternatively, it shouldn't be called path but directory.
("path" makes me think of PATH and LD_LIBRARY_PATH etc. which are lists of directories)

kdwsdl2cpp/common/fileprovider.cpp
caspermeijn: Here I matched the other debug output. When the file will be downloaded, then it will be displayed as well. 
The URL is shown on the debug line before this. It will look something like this:
````
importing schema at http://docs.oasis-open.org/wsrf/bf-2.xsd
Using import path '/home/casper/QtProjects/onvifviewer/3rdparty/wsdl/docs.oasis-open.org//wsrf/bf-2.xsd'
````

**Comment by caspermeijn** [comment]
> And yes there are unittests that use kdwsdl2cpp -- everything under unittests that contains a .wsdl file ;)

I will add an test for this soon.

> Another thought: how about "the directory where the wsdl file is" as a default import path? Then people can install the xsd files locally and they will be found automatically without the need to even specify -import-path (which is of course still useful in case people want to add more search paths, e.g. because some .xsd files are shared between .wsdl files which are not in the same directory)

What I did was downloading all files including the wsdl to the import path (see [this repo](https://gitlab.com/caspermeijn/onvifviewer/tree/kdsoap_import_path/3rdparty/wsdl)). This way I have a uniform way of saving the files. The directory structure is automatically generated using wget (for example `wget -r http://www.w3.org/2005/05/xmlmime`) and this can easily be [scripted](https://gitlab.com/caspermeijn/onvifviewer/blob/kdsoap_import_path/3rdparty/wsdl/download.sh). With my directory structure it is not logical to make the wsdl-directory the default import path.

> And then when a file is NOT found locally, we can show a useful message like "download this file into [this subdirectory] and it will be found locally next time".

Would it be a good idea to add another command line option to disable downloading? When this is enabled and a file is not found in the import path, then the the message is shown and the command exits with an error code. This way you can ensure that all files are available in the import path and you can give a clear explanation for downloading the file to the import path. This option could also be handy for the tests.

**Comment by dfaure-kdab** [comment]
Yes, I meant using wget, but specifically in the directory of the wsdl file. So you get www.w3.org/2005/05/xmlmime indeed, but if kdwsdl2cpp would automatically look in the directory of the wsdl file when looking for www.w3.org, then the developer wouldn't need to specify a search path at all.

The search path option remains useful, but only for more complex cases (of sharing files between wsdl).

I like the idea of an option that forbids any downloading, indeed.

File: kdwsdl2cpp/common/fileprovider.cpp [hunk 4]
```diff
@@ -63,6 +65,17 @@ bool FileProvider::get( const QUrl &url, QString &target )
       return true;
   }
 
+  QStringList importPathList = Settings::self()->importPathList();
+  Q_FOREACH( QString importPath, importPathList ) {
```
dfaure-kdab: Q_FOREACH (const QString &importPath, importPathList) {

kdwsdl2cpp/src/main.cpp
```diff
@@ -125,6 +130,13 @@ int main(int argc, char **argv)
             }
         } else if (opt == QLatin1String("-keep-unused-types")) {
             keepUnusedTypes = true;
+        } else if (opt == QLatin1String("-import-path")) {
+            ++arg;
+            if (!argv[arg]) {
+                showHelp(argv[0]);
+                return 1;
+            }
+            importPathList.append(argv[arg]);
```
dfaure-kdab: append(QFile::decodeName(argv[arg])), to handle correctly non-ascii characters [on Unix at least...].

caspermeijn:
These new commits should address the earlier comment.

kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -63,6 +65,17 @@ bool FileProvider::get( const QUrl &url, QString &target )
       return true;
   }
 
+  QStringList importPathList = Settings::self()->importPathList();
+  Q_FOREACH (QString importPath, importPathList) {
```
dfaure-kdab: const QString &

kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -63,6 +65,17 @@ bool FileProvider::get( const QUrl &url, QString &target )
       return true;
   }
 
+  QStringList importPathList = Settings::self()->importPathList();
```
dfaure-kdab: const QStringList ...
(so the Q_FOREACH can be ported to c++11 range-for more easily later)

kdwsdl2cpp/common/fileprovider.cpp
dfaure-kdab: Yes I got confused by what's in "path", sorry.

kdwsdl2cpp/src/settings.cpp
```diff
@@ -205,3 +205,18 @@ void Settings::setNameSpace(const QString &ns)
...
+QStringList Settings::importPathList() const
+{
+    return mImportPathList;
+}
+
+void Settings::setImportPathList(const QStringList &importPathList)
+{
+    mImportPathList = importPathList;
+    if (mWsdlUrl.isLocalFile()) {
+        QFileInfo wsdlFileInfo(mWsdlUrl.path());
```
dfaure-kdab: .toLocalFile() rather than .path(), to avoid problems on Windows.

kdwsdl2cpp/src/settings.cpp
```diff
@@ -205,3 +205,18 @@ void Settings::setNameSpace(const QString &ns)
...
+        QString wsdlDirPath = wsdlFileInfo.dir().path();
```
dfaure-kdab: wsdlFileInfo.absolutePath() would be faster, no need for the intermediate QDir object.

caspermeijn:
No worries, the shown date is the original commit date. I only pushed the changes yesterday.

caspermeijn:
@dfaure-kdab Did you notice the amended commits?

caspermeijn:
Thanks!

PR #135, opened 2018-08-09
KDSoapServer: Fix for Qt without SSL compiled in

caspermeijn:
The compilation of KDSoap was failing if Qt is build without SSL
enabled. The define QT_NO_OPENSSL is set in the
QtNetwork/qtnetwork-config.h header, which is not included before the
define is used. QTcpSocket will load header for us and was already
always included (either directly or via QSslSocket).

### DIFF
--- a/src/KDSoapServer/KDSoapServerSocket_p.h
+++ b/src/KDSoapServer/KDSoapServerSocket_p.h
@@ -25,10 +25,9 @@
 
 #include <QtGlobal>
 
+#include <QTcpSocket> //may define QT_NO_OPENSSL
 #ifndef QT_NO_OPENSSL
 #include <QSslSocket>
-#else
-#include <QTcpSocket>
 #endif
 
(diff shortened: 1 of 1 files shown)


dfaure-kdab:
Looks good, but are you sure the krazy comment is needed?

I thought krazy only checked "own header first", which this isn't breaking.

src/KDSoapServer/KDSoapServerSocket_p.h
```diff
@@ -25,10 +25,9 @@
 
 #include <QtGlobal>
 
+#include <QTcpSocket> //krazy:exclude=includes must come first to define QT_NO_OPENSSL
```
winterz: David is correct. you don't need to say krazy:exclude here.  the comment should be "may define QT_NO_OPENSSL"

caspermeijn:
Thanks

PR #140, opened 2018-10-02
Update KDSoapServerSocket_p.h

DRFrederik:
On my platform, using Qt 5.9.6, QT_NO_OPENSSL is defined in qtnetwork-config.h
In KDSoapServerSocket_p.h line 29, QT_NO_OPENSSL is not defined which leads to the inclusion of QSslSocket.h.
QSslSocket.h includes qtnetwork-config.h indirectly, now QT_NO_OPENSSL is defined, KDSoapServerSocket does not inherit from the already included QSslSocket but from QTcpSocket which was not included ...

qtnetwork-config.h must be included before using QT_NO_OPENSSL.

### DIFF
--- a/src/KDSoapServer/KDSoapServerSocket_p.h
+++ b/src/KDSoapServer/KDSoapServerSocket_p.h
@@ -24,6 +24,7 @@
 #define KDSOAPSERVERSOCKET_P_H
 
 #include <QtGlobal>
+#include <qtnetwork-config.h>
 
 #include <QTcpSocket> //may define QT_NO_OPENSSL
 #ifndef QT_NO_OPENSSL


DRFrederik:
QSslSocket and QTcpSocket do indeed include qtnetwork-config.h indirectly.
The point is that on line 29 of KDSoapServerSocket_p.h, ifndef QT_NO_OPENSSL is used. At this point QT_NO_OPENSSL is not defined, this eveluates to QSslSocket being included which includes qtnetwork-config.h indirectly. qtnetwork-config.h now defines QT_NO_OPENSSL. At line 44-45 another ifndef QT_NO_OPENSSL now evaluates to false because QT_NO_OPENSSL is defined, this causes KDSoapServerSocket to inherit from QTcpSocket while it was not included in the first ifndef ...

caspermeijn:
Hi @DRFrederik, I recently make a change in the same file because I noticed this problem as well. See https://github.com/KDAB/KDSoap/commit/8c6ba546f11de1272cfe419479b2d76dcf11e2c2. Are you building using the latest master of KDSoap?

DRFrederik:
No, I was using the latest release, 1.7.0.
Thanks for pointing this out.

PR #144, opened 2019-01-03
WS-UsernameToken

caspermeijn:
This pull request has some cleanup commits and adds support for WS-UsernameToken. WS-UsernameToken is enabled by setting KDSoapAuthentication::useWSUsernameToken, then a fresh authentication header is added to all requests.

For the unit-test I added KDSoapAuthentication::setOverrideWSUsernameCreatedTime and KDSoapAuthentication::setOverrideWSUsernameNonce to make it reproducible. Do you know a better way of handling this? I think that this way is OK, but it doesn't feel right.

Is it possible to make this feature detectable from CMake, so that my application can fallback to its own implementation for older versions of KDSoap? I am thinking of a CMake variable or version number bump. I don't know what the standard way of doing this is.

### DIFF
--- a/src/KDSoapClient/KDSoapAuthentication.cpp
+++ b/src/KDSoapClient/KDSoapAuthentication.cpp
@@ -139,12 +139,8 @@ bool KDSoapAuthentication::hasWSUsernameTokenHeader() const
     return hasAuth() && d->useWSUsernameToken;
 }
 
-void KDSoapAuthentication::writeWSUsernameTokenHeader(KDSoapNamespacePrefixes &namespacePrefixes, QXmlStreamWriter &writer, const QString &messageNamespace, bool forceQualified) const
+void KDSoapAuthentication::writeWSUsernameTokenHeader(QXmlStreamWriter &writer) const
 {
-    Q_UNUSED(namespacePrefixes);
-    Q_UNUSED(messageNamespace);
-    Q_UNUSED(forceQualified);
-
     if (!hasAuth()) {
         return;
     }
--- a/src/KDSoapClient/KDSoapAuthentication.cpp
+++ b/src/KDSoapClient/KDSoapAuthentication.cpp
@@ -1,5 +1,6 @@
 /****************************************************************************
 ** Copyright (C) 2010-2018 Klaralvdalens Datakonsult AB, a KDAB Group company, info@kdab.com.
+** Copyright (C) 2019 Casper Meijn  <casper@meijn.net>
 ** All rights reserved.
 **
 ** This file is part of the KD Soap library.
@@ -21,15 +22,26 @@
 **
 **********************************************************************/
 #include "KDSoapAuthentication.h"
-#include <QNetworkReply>
-#include <QDebug>
 #include <QAuthenticator>
+#include <QCryptographicHash>
+#include <QDateTime>
+#include <QDebug>
+#include <QNetworkReply>
+#include "KDSoapNamespacePrefixes_p.h"
+#include "KDSoapNamespaceManager.h"
 
 class KDSoapAuthentication::Private
 {
 public:
+    Private() :
+        useWSUsernameToken(false)
+    {}
+
     QString user;
     QString password;
+    bool useWSUsernameToken;
+    QDateTime overrideWSUsernameCreatedTime;
+    QByteArray overrideWSUsernameNonce;
 };
 
 KDSoapAuthentication::KDSoapAuthentication()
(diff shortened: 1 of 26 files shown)


uteichert:
Hi Caspar,

great that you implemented WS-UsernameToken! I'm in the process to do some Onvif-related stuff as well, so this came *exactly* to the right time, so kudos to you!
I am not associated with KDAB in any way, so this comment is not in any way "official" and is not a formal review at all, but as I was greedy, I merged your commits in my KDSoap fork, but there are some problems with it:

* the commit 993b9a2994413fcef3ec5ac8051b3e1d9ccbe2cd (that's the one you titled with "Remove duplicate code") breaks the thread handling of KDSoap. As soon as you move a client request to another thread (with QObject::moveToThread) you will get warnings like:
QObject: Cannot create children for a parent that is in a different thread.
(Parent is QNetworkAccessManager(0x279b48), parent's thread is QThread(0x27b774), current thread is KDSoapClientThread(0x247ab4)
After reverting that commit everything was working as before, so the code you deleted did something different, I suppose.

* in src/KDSoapClient/KDSoapAuthentication.cpp you are missing some Qt-includes
* also in src/KDSoapClient/KDSoapAuthentication.cpp, you are using a C++11 feature - which would be the first and only usage of C++11 inside KDSoap. I think this should at least be discussed and I would vote for either doing something like:
Private() : useWSUsernameToken(false) {}
instead of
bool useWSUsernameToken = false;
or bump the version number - perhaps to 2.0 - and list this as breaking change.

After fixing these minor issues, I'll now going to try out your feature ;-)

caspermeijn:
> Hi Caspar,
> 
> great that you implemented WS-UsernameToken! I'm in the process to do some Onvif-related stuff as well, so this came _exactly_ to the right time, so kudos to you!
> I am not associated with KDAB in any way, so this comment is not in any way "official" and is not a formal review at all, but as I was greedy, I merged your commits in my KDSoap fork, but there are some problems with it:

Thanks for you reaction, this helps a lot :-)

> 
>     * the commit [993b9a2](https://github.com/KDAB/KDSoap/commit/993b9a2994413fcef3ec5ac8051b3e1d9ccbe2cd) (that's the one you titled with "Remove duplicate code") breaks the thread handling of KDSoap. As soon as you move a client request to another thread (with QObject::moveToThread) you will get warnings like:
>       QObject: Cannot create children for a parent that is in a different thread.
>       (Parent is QNetworkAccessManager(0x279b48), parent's thread is QThread(0x27b774), current thread is KDSoapClientThread(0x247ab4)
>       After reverting that commit everything was working as before, so the code you deleted did something different, I suppose.

I remove this commit from the pull request, as I now see that it was wrong. I wanted to de-duplicate the code as I wanted to adjust the code, however in the final version this was not needed. So the commit was not needed for the code to work.

> 
>     * in src/KDSoapClient/KDSoapAuthentication.cpp you are missing some Qt-includes

These includes were only missing for Qt4, I added them.

> 
>     * also in src/KDSoapClient/KDSoapAuthentication.cpp, you are using a C++11 feature - which would be the first and only usage of C++11 inside KDSoap. I think this should at least be discussed and I would vote for either doing something like:
>       Private() : useWSUsernameToken(false) {}
>       instead of
>       bool useWSUsernameToken = false;
>       or bump the version number - perhaps to 2.0 - and list this as breaking change.
> 

You are right, these C++11 feature are so normal to me. I removed that code. 

However I noticed the CMakeLists states: set(CMAKE_CXX_STANDARD 11)

>     * You are using QT5-only features

I fixed those.

> 
> 
> After fixing these minor issues, I'll now going to try out your feature ;-)

jpost1:
It would be convenient to add the time offset to KDSoapAuthentication. 

auto systemtime = QDateTime::currentDateTimeUtc();
qint64 timeoffset = systemtime.msecsTo(devicetime_utc);

class KDSoapAuthentication {
....
void setTimeOffsetMSecs(quint64 value);
....
}
void KDSoapAuthentication::writeWSUsernameTokenHeader(KDSoapNamespacePrefixes &namespacePrefixes, QXmlStreamWriter &writer, const QString &messageNamespace, bool forceQualified) const
{
  .....
    QDateTime time = QDateTime::currentDateTimeUtc();
    time = time.addMSecs(d->timeOffsetMSecs);
....
}

src/KDSoapClient/KDSoapAuthentication.cpp
```diff
@@ -91,3 +133,57 @@ void KDSoapAuthentication::handleAuthenticationRequired(QNetworkReply *reply, QA
...
 }
+
+bool KDSoapAuthentication::hasWSUsernameTokenHeader() const
+{
+    return hasAuth() && d->useWSUsernameToken;
+}
+
+void KDSoapAuthentication::writeWSUsernameTokenHeader(KDSoapNamespacePrefixes &namespacePrefixes, QXmlStreamWriter &writer, const QString &messageNamespace, bool forceQualified) const
+{
+    Q_UNUSED(namespacePrefixes);
```
dfaure-kdab: Why have those arguments then, if they are unused? Can't they be removed?

caspermeijn: I used the same function signature as KDSoapMessageAddressingProperties::writeMessageAddressingProperties, however that is not actually necessary.

caspermeijn:
> Very nice! Can you add a link to the corresponding spec, in the commit log or better, in KDSoapAuthentication.h?

Added it to KDSoapAuthentication.h

dfaure-kdab:
Yes, we try to keep Qt4 compatibility, but C++11 usage is fine now.

PR #145, opened 2019-01-12
Re-enable support for CMake older than 3.1

rokm:
Commit a387ff19da8166735dd69edbec1478754f8afe41 enables C++11 support, but by doing so also bumps the required CMake version to 3.1. This prevents building of KDSoap on CentOS/RHEL 7, which is still using CMake 2.8.12.

This PR adds a fall-back way of enabling C++11 for gcc with CMake < 3.1, and relaxes the required CMake version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@
 #  Build the test harness.
 #  Default=false
 #
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 2.8.7)
 cmake_policy(SET CMP0020 NEW)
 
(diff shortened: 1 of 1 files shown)


dfaure-kdab:
Oh, I had no idea there were still distros without CMake 3.1 out there.

OK... for now.

dfaure-kdab:
BTW if KDSoap requires a newer cmake at some point in the future, you can also install an overlay for CentOS/RHEL 7 which contains cmake 3.x.

I did that in the past, and IIRC that's as simple as

    yum -y install epel-release
    yum install cmake3

(source: https://www.heliocastro.info/?p=238 in the comments)

PR #146, opened 2019-01-12
Support qname

caspermeijn:
These change add support for sending and receiving qnames. It adds a new KDQName class, this is used to represent the xml qname and the reader and writer are extended to support it.

I am not sure whether this will break existing code, as the soap qname are now represented as KDQName instead of QString. But qnames couldn't be fully used in the past, as the xml namespaces were not correctly passed to the reader/writer.

I am not sure whether this is the simplest solution. I am open for suggestion for simplifying it.

### DIFF
--- a/doc/CHANGES_1_8.txt
+++ b/doc/CHANGES_1_8.txt
@@ -30,6 +30,14 @@ Server-side:
 
 WSDL parser / code generator changes, applying to both client and server side:
 ================================================================
+* Source incompatible change: all deserialize() functions now require a KDSoapValue instead of a 
+    QVariant. If you use a deserialize(QVariant) function, you need to port your code to use 
+    KDSoapValue::setValue(QVariant) before deserialize()
+* Source incompatible change: all serialize() functions now return a KDSoapValue instead of a 
+    QVariant. If you use a QVariant serialize() function, you need to port your code to use 
+    QVariant KDSoapValue::value() after serialize()
+* Source incompatible change: xs:QName is now respresented by KDQName instead of QString, which 
+    allows the namespace to be extracted. The old behaviour is available via KDQName::qname().
 * Fix double-handling of empty elements
 * Fix fault elements being generated in the wrong namespace, must be SOAP-ENV:Fault (github issue #81).
 * Added import-path argument for setting the local path to get (otherwise downloaded) files from.
@@ -41,3 +49,5 @@ WSDL parser / code generator changes, applying to both client and server side:
 * Added "hasValueFor{MemberName}()" accessor function, for optional elements
 * Generated services now include soapVersion() and endpoint() accessors to match the setSoapVersion(...) and setEndpoint(...) mutators
 * Added support for generating messages for WSDL files without services or bindings
+* KDSoapValue now stores the namespace declarations during parsing of a message and writes 
+    namespace declarations during sending of a message
--- a/src/KDSoapClient/KDSoapValue.cpp
+++ b/src/KDSoapClient/KDSoapValue.cpp
@@ -367,6 +367,20 @@ QString KDSoapValue::type() const
     return d->m_typeName;
 }
 
+KDSoapValueList KDSoapValue::split() const
+{
+    KDSoapValueList valueList;
+    if (value().toString().trimmed().isEmpty()) return valueList;
+
+    const QStringList list = value().toString().split(QLatin1Char(' '));
+    for (int i = 0; i < list.count(); ++i) {
+        KDSoapValue value(*this);
+        value.setValue(list.at(i));
+        valueList << value;
+    }
+    return valueList;
+}
+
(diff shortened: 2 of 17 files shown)


src/KDSoapClient/KDSoapValue.cpp
```diff
@@ -367,6 +367,20 @@ QString KDSoapValue::type() const
     return d->m_typeName;
 }
 
+KDSoapValueList KDSoapValue::split() const
+{
+    KDSoapValueList valueList;
+    if (value().toString().trimmed().isEmpty()) return valueList;
```
dfaure-kdab: This line seems slow and unnecessary? Well, the trimmed() makes a difference, but I guess skipping empty list items in the for loop below would solve this too (more generically, even)?

src/KDSoapClient/KDSoapValue.cpp
```diff
@@ -367,6 +367,20 @@ QString KDSoapValue::type() const
...
+
+    const QStringList list = value().toString().split(QLatin1Char(' '));
+    for (int i = 0; i < list.count(); ++i) {
+        KDSoapValue value(*this);
+        value.setValue(list.at(i));
+        valueList << value;
```
dfaure-kdab: valueList.reserve(list.count()) should be done before the for loop, to minimize re-allocations.

src/KDSoapClient/KDSoapValue.cpp
```diff
@@ -285,6 +290,10 @@ void KDSoapValue::writeElementContents(KDSoapNamespacePrefixes &namespacePrefixe
 {
     const QVariant value = this->value();
 
+    foreach (QXmlStreamNamespaceDeclaration decl, d->m_namespaceDeclarations) {
```
dfaure-kdab: const & missing?

caspermeijn:
I believe there is no example of qname in the current tests. The best example I know is WS-Discovery. I opened another pull request, where I add a test for that. I will modify that test to show qname support as soon as #161 lands in master.

dfaure-kdab:
#161 is in master now :-)

caspermeijn:
I fixed a problem with namespaces defined by a parent element and added a unit test.

doc/CHANGES_1_8.txt
```diff
@@ -41,3 +41,4 @@ WSDL parser / code generator changes, applying to both client and server side:
 * Added "hasValueFor{MemberName}()" accessor function, for optional elements
 * Generated services now include soapVersion() and endpoint() accessors to match the setSoapVersion(...) and setEndpoint(...) mutators
 * Added support for generating messages for WSDL files without services or bindings
+* Added proper support for QNames, which will allow the namespace to be extracted
```
dfaure-kdab: This needs to be much more detailed so people know what to expect. I'd mention that the API for serialize/deserialize has changed to take/return KDSoapValue instead of QVariant, for one thing. Anyone using that will have to port their code.

And then the generated code changed API when using xs:QName -- but I guess before your earlier commit this wouldn't compile at all?

This changelog paragraph should start with "Source incompatible change:" so people know that this might happen, like CHANGES_1_5.txt does.

unittests/wsdl_document/test_wsdl_document.cpp
```diff
@@ -1060,7 +1060,7 @@ void WsdlDocumentTest::testSendTelegram()
 
     // Check the request as received by the server.
     const QByteArray expectedRequestXml =
-        QByteArray(xmlBegin) + "<TelegramRequest " + xmlNamespaces + ">"
+        QByteArray(xmlBegin) + "<TelegramRequest " + xmlNamespaces + " xmlns:n1=\"http://www.kdab.com/xml/MyWsdl/\">"
```
dfaure-kdab: (what's strange with this one is that n1 isn't actually used... doesn't hurt though, I guess)

doc/CHANGES_1_8.txt
caspermeijn: I replaced the changelog commit in this branch

doc/CHANGES_1_8.txt
```diff
@@ -30,6 +30,14 @@ Server-side:
 
 WSDL parser / code generator changes, applying to both client and server side:
 ================================================================
+* Source incompatible change: all deserialize() functions now require a KDSoapValue instead of a 
+    QVariant. If you use a deserialize(QVariant) function, you need to port your code to use 
+    KDSoapValue::setValue(QVariant) before deserialize()
+* Source incompatible change: all serialize() functions now return a KDSoapValue instead of a 
+    QVariant. If you use a QVariant serialize() function, you need to port your code to use 
+    QVariant KDSoapValue::value() after serialize()
+* Source incompatible change: xs:QName is now respresented by KDQName instead of QString, which
```
dfaure-kdab: respresented -> represented, I'll fix it after merge

PR #147, opened 2019-01-13
WS-Addressing reader

caspermeijn:
This pull request adds support for reading WS-Addressing headers for incoming messages. The WS-Addressing headers are placed in KDSoapMessageAddressingProperties, so that the application can use them easily.

It also has a patch for the WS-Addressing writer which doesn't requires destination and source as the [spec](https://www.w3.org/TR/2006/REC-ws-addr-core-20060509/) says they are optional.


dfaure-kdab:
Very nice. All that's missing is an entry in doc/CHANGES_1_8.txt :-)
I'll add it after merging this.

PR #149, opened 2019-02-01
Fix if a referenced type name has an extra space on the ends. 

ryanenzo:
E.g. allow the type name "xs:string " to work. For when you can't fix the source WSDL.


dfaure-kdab:
Not sure the XML Schema spec says this should work, but well, it can't hurt.

PR #150, opened 2019-02-01
Cache downloaded files.

ryanenzo:
Download files only once. E.g. If a WSDL imports A.xsd and B.xsd, and they both import Core.xsd, only download Core.xsd once.
Also make indentation consistent at 4 spaces.


dfaure-kdab:
Please don't change the indentation. I know it's a bit weird, but some of the subdirs in kdwsdl2cpp actually come from libkode (http://github.com/cornelius/kode) which uses 2 spaces indentation.
So we need to keep it that way to be able to sync the two repos.

I actually have work in progress to undo some accidental reformatting in the schema and wsdl subdirs, so that it's comparable again to libkode...

Can you also merge the two commits and force-push, so that the diff is readable? Thanks.

PR #151, opened 2019-02-01
- Add support for setting request headers to KDSoapJob

ryanenzo:
Also rename "KDSoapJob::returnHeaders" to "KDSoapJob::replyHeaders" and actually provide an implementation

### DIFF
--- a/src/KDSoapClient/KDSoapJob.h
+++ b/src/KDSoapClient/KDSoapJob.h
@@ -87,6 +87,8 @@ public:
     /**
      * Sets request headers to be sent to the SOAP server. These are sent in addition
      * to the persistent headers set via the client interface.
+     *
+     * \since 1.8
      */
     void setRequestHeaders(const KDSoapHeaders &headers);
 
@@ -109,6 +111,8 @@ public:
     /**
      * Returns the reply headers received from the SOAP server once the request was completed.
      * Only valid once the request is completed and finished() was emitted.
+     *
+     * \since 1.8
      */
     KDSoapHeaders replyHeaders() const;
 
--- a/src/KDSoapClient/KDSoapJob.h
+++ b/src/KDSoapClient/KDSoapJob.h
@@ -79,6 +79,18 @@ public:
     ~KDSoapJob();
 
     /**
+     * Returns the reply headers received from the SOAP server once the request was completed.
+     * Only valid once the request is completed and finished() was emitted.
+     */
+    KDSoapHeaders requestHeaders() const;
+
+    /**
+     * Sets request headers to be sent to the SOAP server. These are sent in addition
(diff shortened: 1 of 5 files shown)


src/KDSoapClient/KDSoapJob.h
```diff
@@ -98,7 +110,7 @@ class KDSOAP_EXPORT KDSoapJob : public QObject
      * Returns the reply headers received from the SOAP server once the request was completed.
      * Only valid once the request is completed and finished() was emitted.
      */
-    KDSoapHeaders returnHeaders() const;
+    KDSoapHeaders replyHeaders() const;
```
dfaure-kdab: You can't rename a public method, that's source incompatible.

If you want to rename this, keep the old method, mark it as deprecated (like Qt does), add the new method next to it (with "\since 1.8").

(Note I think this name initially comes from KDSoapPendingCall, so initially I was going to say "it should remain consistent", but actually... not really. For one call, there's a return value, while here it's about reply to the request, OK.)

src/KDSoapClient/KDSoapJob.h
```diff
@@ -78,6 +78,18 @@ class KDSOAP_EXPORT KDSoapJob : public QObject
      */
     ~KDSoapJob();
 
+    /**
+     * Returns the reply headers received from the SOAP server once the request was completed.
+     * Only valid once the request is completed and finished() was emitted.
```
dfaure-kdab: \since 1.8 missing

src/KDSoapClient/KDSoapJob.h
```diff
@@ -78,6 +78,18 @@ class KDSOAP_EXPORT KDSoapJob : public QObject
...
+     */
+    KDSoapHeaders requestHeaders() const;
+
+    /**
+     * Sets request headers to be sent to the SOAP server. These are sent in addition
+     * to the persistent headers set via the client interface.
+     */
+    void setRequestHeaders(const KDSoapHeaders &headers);
```
dfaure-kdab: \since 1.8 missing

src/KDSoapClient/KDSoapJob.h
ryanenzo: Normally I would agree about not renaming public methods ... but the implementation was missing. If you actually tried to use "returnHeaders()" you would get a linker error.

dfaure-kdab: LOL that's interesting. Whoops, looks like I went a bit fast on that day. Then I fully agree with this change :)

dfaure-kdab:
Can you add an entry in doc/CHANGES_1_8.txt for this change and all others? Thanks.

PR #152, opened 2019-02-01
Several changes

ryanenzo:
- Add ability to generate both .h and .cpp file in one run of kdwsdl2cpp 
- Add support for namespace-mapping to control class-name generation 
  (e.g. to generate "MYPREFIX__MyClass" instead of "TNS__MyClass")
- Add endpoint() and soapVersion() accessors in generated Wsdl Service classes
- Fixes to prevent generation of some invalid C++

### DIFF
--- a/kdwsdl2cpp/libkode/style.cpp
+++ b/kdwsdl2cpp/libkode/style.cpp
@@ -83,6 +83,9 @@ QString Style::lowerFirst( const QString &str )
 
 QString Style::makeIdentifier( const QString &str )
 {
+  if ( str.isEmpty() )
+    return str;
+
   QString identifier = str;
   identifier.replace( "-", "_" );
   identifier.replace( ".", "_" );
--- a/kdwsdl2cpp/libkode/style.cpp
+++ b/kdwsdl2cpp/libkode/style.cpp
@@ -83,8 +83,7 @@ QString Style::lowerFirst( const QString &str )
 
 QString Style::makeIdentifier( const QString &str )
 {
-  if ( str.isEmpty() )
-    return str;
+  Q_ASSERT(!str.isEmpty());
 
   QString identifier = str;
   identifier.replace( "-", "_" );
--- a/kdwsdl2cpp/src/compiler.cpp
+++ b/kdwsdl2cpp/src/compiler.cpp
@@ -103,7 +103,6 @@ void Compiler::parse(const QDomElement &element)
 
         definitions.fixUpDefinitions(/*&context, element*/);
 
-        KODE::Code::setDefaultIndentation(4);
 
         WSDL wsdl;
         wsdl.setDefinitions(definitions);
@@ -116,7 +115,19 @@ void Compiler::parse(const QDomElement &element)
             QCoreApplication::exit(4);
         } else {
             KWSDL::Creator creator;
-            creator.create(converter.classes());
+            creator.setOutputDirectory(Settings::self()->outputDirectory());
+            creator.setSourceFile(Settings::self()->wsdlFileName());
+            creator.setHeaderFileName(Settings::self()->headerFileName());
+            creator.setImplementationFileName(Settings::self()->implementationFileName());
+
+            creator.setClasses(converter.classes());
+            if (Settings::self()->generateHeader()) {
+                creator.createHeader();
+            }
+            if (Settings::self()->generateImplementation()) {
+                creator.createImplementation();
+            }
+
             QCoreApplication::exit(0);
         }
     } else {
--- a/kdwsdl2cpp/src/converter_complextype.cpp
+++ b/kdwsdl2cpp/src/converter_complextype.cpp
@@ -644,7 +644,9 @@ void Converter::createComplexTypeSerializer(KODE::Class &newClass, const XSD::Co
         bool first = true;
         Q_FOREACH (const XSD::Attribute &attribute, attributes) {
             const QString attrName = attribute.name();
-            if (attrName.isEmpty()) { continue; }
+            if (attrName.isEmpty()) {
+                continue;
+            }
             const QString variableName = QLatin1String("d_ptr->") + KODE::MemberVariable::memberVariableName(attrName);
 
             demarshalCode.addBlock(demarshalNameTest(attribute.type(), attrName, &first));
--- a/kdwsdl2cpp/src/converter_complextype.cpp
+++ b/kdwsdl2cpp/src/converter_complextype.cpp
@@ -192,7 +192,8 @@ void Converter::convertComplexType(const XSD::ComplexType *type)
 
         typeName = mTypeMap.localType(attribute.type());
         if (typeName.isEmpty()) {
-            qDebug() << "ERROR: attribute with unknown type:" << attribute.name() << attribute.type() << "in" << typeName;
+            qWarning() << "ERROR: attribute with unknown type:" << attribute.name() << attribute.type() << "in" << typeName;
+            continue;
         }
         inputTypeName = mTypeMap.localInputType(attribute.type(), QName());
         //qDebug() << "Attribute" << attribute.name();
@@ -643,6 +644,7 @@ void Converter::createComplexTypeSerializer(KODE::Class &newClass, const XSD::Co
         bool first = true;
         Q_FOREACH (const XSD::Attribute &attribute, attributes) {
             const QString attrName = attribute.name();
+            if (attrName.isEmpty()) { continue; }
             const QString variableName = QLatin1String("d_ptr->") + KODE::MemberVariable::memberVariableName(attrName);
 
             demarshalCode.addBlock(demarshalNameTest(attribute.type(), attrName, &first));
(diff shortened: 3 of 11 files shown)


kdwsdl2cpp/libkode/style.cpp
```diff
@@ -83,6 +83,9 @@ QString Style::lowerFirst( const QString &str )
 
 QString Style::makeIdentifier( const QString &str )
 {
+  if ( str.isEmpty() )
```
dfaure-kdab: Strange place for this check, "" isn't a valid identifier in the first place... No big deal but it makes me wonder, who's calling this with an empty identifier?

kdwsdl2cpp/src/compiler.cpp
```diff
@@ -103,7 +103,6 @@ void Compiler::parse(const QDomElement &element)
 
         definitions.fixUpDefinitions(/*&context, element*/);
 
-        KODE::Code::setDefaultIndentation(4);
```
dfaure-kdab: Why was this removed?

kdwsdl2cpp/src/converter_complextype.cpp
```diff
@@ -643,6 +644,7 @@ void Converter::createComplexTypeSerializer(KODE::Class &newClass, const XSD::Co
         bool first = true;
         Q_FOREACH (const XSD::Attribute &attribute, attributes) {
             const QString attrName = attribute.name();
+            if (attrName.isEmpty()) { continue; }
```
dfaure-kdab: On 3 lines please.

kdwsdl2cpp/src/creator.h
```diff
@@ -21,16 +21,30 @@
...
 #include <libkode/class.h>
+#include <libkode/file.h>
+#include <libkode/license.h>
+#include <libkode/printer.h>
 
 namespace KWSDL
 {
 
-class Creator
+class Creator : private KODE::Printer
```
dfaure-kdab: I would find it more readable to have the printer as a member variable.

kdwsdl2cpp/src/main.cpp
```diff
@@ -123,6 +147,40 @@ int main(int argc, char **argv)
...
+            if (!argv[arg]) {
+                showHelp(argv[0]);
+                return 1;
+            }
+            QString mapping = argv[arg];
+            if (mapping.startsWith('@')) {
+                QString mappingFileName = mapping.mid(1);
+                QFile file(mappingFileName);
+                if (!file.open(QIODevice::ReadOnly)) {
+                    fprintf(stderr, "Error reading %s: %s\n", mappingFileName.toLatin1().data(), file.errorString().toLatin1().data());
```
dfaure-kdab: `QFile::encodeName(mappingFileName).constData()` would be more correct (in case of non-latin1 chars in the path)

kdwsdl2cpp/src/main.cpp
```diff
@@ -123,6 +147,40 @@ int main(int argc, char **argv)
...
+                    showHelp(argv[0]);
+                    return 1;
+                }
+
+                while(!file.atEnd()) {
```
dfaure-kdab: missing space after `while`

kdwsdl2cpp/src/main.cpp
```diff
@@ -123,6 +147,40 @@ int main(int argc, char **argv)
...
+                    QString mapping = file.readLine().trimmed();
+                    if (mapping.startsWith("#")) { continue; }
```
dfaure-kdab: `'#'` and indent on 3 lines

kdwsdl2cpp/src/main.cpp
```diff
@@ -163,19 +221,52 @@ int main(int argc, char **argv)
...
+    }
+
+
+    if (both) {
+        Settings::self()->setGenerateHeader(true);
+        Settings::self()->setGenerateImplementation(true);
+        Settings::self()->setHeaderFileName(baseFile + ".h");
+        Settings::self()->setImplementationFileName(baseFile + ".cpp");
+    }
+    else if (impl) {
```
dfaure-kdab: join with previous line

kdwsdl2cpp/src/main.cpp
```diff
@@ -163,19 +221,52 @@ int main(int argc, char **argv)
...
+        Settings::self()->setGenerateHeader(false);
+        Settings::self()->setGenerateImplementation(true);
+        Settings::self()->setHeaderFileName(headerFile);
+        Settings::self()->setImplementationFileName(outputFile.fileName());
+    } 
+    else {
```
dfaure-kdab: join with previous line

kdwsdl2cpp/src/main.cpp
```diff
@@ -163,19 +221,52 @@ int main(int argc, char **argv)
...
     Settings::self()->setExportDeclaration(exportMacro);
     Settings::self()->setNameSpace(nameSpace);
+    Settings::self()->setNamespaceMapping(nsmapping);
     Settings::self()->setOptionalElementType(optionalElementType);
     Settings::self()->setKeepUnusedTypes(keepUnusedTypes);
     Settings::self()->setImportPathList(importPathList);
     Settings::self()->setUseLocalFilesOnly(useLocalFilesOnly);
     Settings::self()->setHelpOnMissing(helpOnMissing);
+
+    //fprintf(stderr, "Output dir: %s\n", Settings::self()->outputDirectory().toLatin1().data());
```
dfaure-kdab: I think you can remove all this debugging now

kdwsdl2cpp/src/settings.cpp
```diff
@@ -101,11 +102,17 @@ QString Settings::wsdlFileName() const
...
-    mOutputFileName = outputFileName;
+    mHeaderFileName = headerFileName;
 }
 
+void Settings::setImplementationFileName(const QString &implementationFileName)
+{
+    mImplementationFileName = implementationFileName;
+}
+
+/*
```
dfaure-kdab: kill dead code

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -58,7 +58,7 @@ void KDSoapClientInterface::setSoapVersion(KDSoapClientInterface::SoapVersion ve
     d->m_version = static_cast<KDSoap::SoapVersion>(version);
 }
 
-KDSoapClientInterface::SoapVersion KDSoapClientInterface::soapVersion()
+KDSoapClientInterface::SoapVersion KDSoapClientInterface::soapVersion() const
```
dfaure-kdab: This is unrelated, and I'm pretty sure I saw it in a different review request already, please remove from this commit.

kdwsdl2cpp/src/main.cpp
```diff
@@ -163,19 +221,52 @@ int main(int argc, char **argv)
...
+        Settings::self()->setGenerateHeader(false);
+        Settings::self()->setGenerateImplementation(true);
+        Settings::self()->setHeaderFileName(headerFile);
+        Settings::self()->setImplementationFileName(outputFile.fileName());
+    } 
+    else {
+        Settings::self()->setGenerateHeader(true);
+        Settings::self()->setGenerateImplementation(false);
+        Settings::self()->setHeaderFileName(outputFile.fileName());
+        Settings::self()->setImplementationFileName("UNUSED");
```
dfaure-kdab: Why call it then? To see an obvious error if this is ever used? ;-)

kdwsdl2cpp/src/main.cpp
```diff
@@ -163,19 +221,52 @@ int main(int argc, char **argv)
         return 1;
     }
 
+    // if you're saying "just make the impl-file", you can't
+    //    also say "make both the header and the impl"
+    if (both && impl) {
```
dfaure-kdab: Same if specifying -o and -both, right? This should give an error too.
To do it here you'd need another bool, the alternative would be to test it when parsing -both (outputFile should still be empty) and when parsing -o (both should be false).

kdwsdl2cpp/libkode/style.cpp
ryanenzo: I don't recall the situation, but it was causing a crash. Thus this was added.

kdwsdl2cpp/src/compiler.cpp
ryanenzo: Not removed, just moved into the Creator class.

kdwsdl2cpp/libkode/style.cpp
dfaure-kdab: Yes the fact that this would crash is obvious, given that at(0) in the implementation.

But it's still an invalid use of this method, in my opinion.
Make it a `Q_ASSERT(!str.isEmpty())` for clarity that this is the contract, but then you'll still have to fix the caller.

dfaure-kdab: This is still pending.

ryanenzo: Shoudl be done now.

PR #153, opened 2019-02-01
Cache downloads. Prevents redownloading the same imported file twice.

ryanenzo:


kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -94,30 +97,38 @@ bool FileProvider::get( const QUrl &url, QString &target )
...
-  QEventLoop loop;
-  connect(job, SIGNAL(finished()),
-          &loop, SLOT(quit()));
-  loop.exec();
-
-  if (job->error()) {
-      qWarning("Error downloading '%s': %s", url.toEncoded().constData(), qPrintable(job->errorString()));
-      return false;
+  QByteArray data;
+  if (fileProviderCache.contains(url)) {
```
dfaure-kdab: Nitpick: this is a double lookup, better written as
```
QByteArray data = fileProviderCache.value(url);
if (!data.isEmpty()) {
    qDebug...
} else {
```

kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -94,30 +97,38 @@ bool FileProvider::get( const QUrl &url, QString &target )
...
+    loop.exec();
+
+    if (job->error()) {
+        qWarning("Error downloading '%s': %s", url.toEncoded().constData(), qPrintable(job->errorString()));
+        return false;
+    }
+
+    qDebug( "Download successful" );
+    data = job->readAll();
+    fileProviderCache[url] = data;
```
dfaure-kdab: better written as `fileProviderCache.insert(url, data)`

I know, performance doesn't matter here, but it's just better practice. (reference: `Effective STL` book)

kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -94,30 +97,38 @@ bool FileProvider::get( const QUrl &url, QString &target )
...
   }
 
-  const QByteArray data = job->readAll();
   QFile file( mFileName );
   if ( !file.open( QIODevice::WriteOnly ) ) {
-      qDebug( "Unable to create temporary file" );
+      qDebug( qPrintable("Unable to create temporary file: " + file.errorString()) );
```
dfaure-kdab: Easier (and more informative) like this:
```
    qWarning() << "Unable to create" << mFileName << ":" << file.errorString();
```

kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -37,6 +37,9 @@
 #include <unistd.h>
 #endif
 
+QHash<QUrl, QByteArray> fileProviderCache;
```
dfaure-kdab: prepend `static`

kdwsdl2cpp/common/fileprovider.cpp
ryanenzo: This is a QMap, not an STL map. And since this is easier to read, and is not a performance matter, I'm leaving it.

dfaure-kdab: QMap has exactly the same behaviour as an STL map since this is a C++ issue. [] has to create an empty value which is then replaced with the actual value.

Easier to read is a matter of habit. Every time I see [] being used the alarm bells ring in my head because of unwanted detach (when reading) or because of this issue (when writing). value() and insert() solve both these issues.

But yes, OK, I'll accept the commit with [] if you won't change it ;)

kdwsdl2cpp/common/fileprovider.cpp
ryanenzo: That makes an assumption about valid data. Theoretically, the downloaded data could be empty. Using the QMap::find and iterators is the proper way to do this.

dfaure-kdab: You're right, this is a case where the map _could_ contain an empty value. Would likely give a parse error down the line though ;)
OK, let's leave it at that.

kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -94,30 +97,39 @@ bool FileProvider::get( const QUrl &url, QString &target )
...
-  loop.exec();
-
-  if (job->error()) {
-      qWarning("Error downloading '%s': %s", url.toEncoded().constData(), qPrintable(job->errorString()));
-      return false;
+  QByteArray data;
+  const QHash<QUrl, QByteArray>::const_iterator it = fileProviderCache.constFind(url);
+  if (it != fileProviderCache.constEnd()) {
+    data = it.value();
+    qDebug( "Using cached copy" );
```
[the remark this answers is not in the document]
ryanenzo: Removed the qDebug.

PR #158, opened 2019-02-04
Extend KDSOAP_DEBUG functionality

ryanenzo:


src/KDSoapClient/KDSoapPendingCall_p.h
```diff
@@ -33,6 +33,8 @@
 
 class KDSoapValue;
 
+void debugHelper(const QByteArray &data, const QNetworkRequest &request, QNetworkReply *reply);
```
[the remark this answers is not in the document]
ryanenzo: Renamed to maybeDebug, maybeDebugRequest, maybeDebugResponse.

src/KDSoapClient/KDSoapPendingCall.cpp
```diff
@@ -82,29 +82,32 @@ void debugHelper(const QByteArray &data, QList< QNetworkReply::RawHeaderPair > h
...
-void debugHelper(const QByteArray &data, QNetworkReply *reply) {
-    debugHelper(data, reply->rawHeaderPairs());
+static void maybeDebugResponse(const QByteArray &data, QNetworkReply *reply) {
+    maybeDebug(data, reply->rawHeaderPairs());
 }
 
-// Log the HTTP and XML of a requst.
-void debugHelper(const QByteArray &data, const QNetworkRequest &request, QNetworkReply *reply) {
+// Log the HTTP and XML of a request.
+// (not static, because this is used in KDSoapClientInterface)
```
dfaure-kdab: Ah right, this is the public one. Then we are doing a lot of work before realizing that we are not going to output any debug...  This will slow down normal usage.

Can you move the env var check so that it happens first thing in this method?

Ah, it needs to happen in maybeDebugResponse too since that's called directly, so that means extracting it to a separate function called by both.

Sorry for not seeing this earlier - the naming change does help reading all this :-)

ryanenzo: Yeah, I moved the env-var check to the first thing in the "mabyeDebug" functions.

PR #159, opened 2019-02-04
Add "hasValue" accessor function for optional values

ryanenzo:


ryanenzo:
I'm really not sure how to add the tests to the setup you have.

dfaure-kdab:
Ah, that should be an easy one: edit unittests/optionaltype_regular/testregularapi.cpp and call the method there, on `resp`, before and after setting a value.

Please do the same in optionaltype_pointer and optionaltype_boost_optional to make sure it works there too.

ryanenzo:
Okay.

Should we also have a “clearValue{Foo}()” (or “clear{Foo}Value()”)?

I would think it just sets the _nil value to true, and doesn’t assign the actual value, avoiding any possible default-constructor issues.

Regards,

-Ryan

ryanenzo:
Added the changes to the unittests

unittests/optionaltype_boost_optional/testboostapi.cpp
```diff
@@ -34,9 +34,11 @@ void TestBoostApi::test()
 {
     TNS__TestOperationResponse1 resp;
     QCOMPARE(resp.out(), boost::optional<QString>());
+    QCOMPARE(resp.hasValueForOut(), false);
     QString newVal("newval");
     resp.setOut(newVal);
     QCOMPARE(*resp.out(), newVal);
+    QCOMPARE(resp.hasValueForOut(), true);
```
dfaure-kdab: This is normally done with QVERIFY(), but OK, I'll stop nitpicking ;)

dfaure-kdab:
I see the idea with clearValueFor{Foo} but I'd say, let's only add it if there's a need for it.
Normally these values aren't reused between calls, so an if() does the job just fine.

dfaure-kdab:
Commit reverted, it breaks compilation of other tests. Did you try building all tests?

```
KDSoap/build_qt5_cmake/unittests/msexchange_noservice_wsdl/wsdl_Services_noservice.cpp: In member function ‘bool TNS__FreeBusyResponseType::hasValueForResponseMessage() const’:
KDSoap/build_qt5_cmake/unittests/msexchange_noservice_wsdl/wsdl_Services_noservice.cpp:1197:19: error: ‘const class TNS__FreeBusyResponseType::PrivateDPtr’ has no member named ‘mResponseMessage_nil’; did you mean ‘mResponseMessage’?
     return d_ptr->mResponseMessage_nil == false;
                   ^~~~~~~~~~~~~~~~~~~~
[....]
KDSoap/build_qt5_cmake/unittests/msexchange_wsdl/wsdl_Services.cpp:47501:19: error: ‘const class T__AppendToItemFieldType::PrivateDPtr’ has no member named ‘mItem_nil’; did you mean ‘mItem’?
     return d_ptr->mItem_nil == false;
[....]
```

ryanenzo:
Hi David,

I must admit I did not. For example, in groupwise_wsdl, I get these errors, which appear to be unrelated to my change:
groupwise_wsdl\wsdl_groupwise.h(7405): error C2528: '__daylight': pointer to reference is illegal
groupwise_wsdl\wsdl_groupwise.h(7464): error C2528: '__timezone': pointer to reference is illegal
groupwise_wsdl\wsdl_groupwise.h(11157): error C2528: '__timezone': pointer to reference is illegal

I also get these errors without my modifications, so I basically just ran the tests you wanted me to change. Obviously, that was a mistake on my part.

By, the way, I’m using MSVC 2017.

The offending line of codes:
    void setDaylight( const TYPES__TimezoneComponent& _daylight );
    void setTimezone( const QList< TYPES__Timezone >& _timezone );
    void setTimezone( const TYPES__Timezone& _timezone );

It appears as if _daylight and _timezone are Microsoft variables:
    https://docs.microsoft.com/en-us/cpp/c-runtime-library/daylight-dstbias-timezone-and-tzname?view=vs-2017

I added to the top of the generated .h files, and got rid of this error:
#ifdef _daylight
#undef _daylight
#endif
#ifdef _timezone
#undef _timezone
#endif

PR #160, opened 2019-02-07
Generate "hasValueFor{Value}" functions

ryanenzo:
This now works for polymorphic classes as well. Tests have also been expanded for this.  Change the default name of the argument passed to the setter function. (Fix for _daylight and _timezone i MSVC-2017).

Revert the change in location of the "KODE::setDefaultIndentation" (use of .indent() and .unindent() before calling this will use the old indentation size).

### DIFF
--- a/kdwsdl2cpp/src/converter_complextype.cpp
+++ b/kdwsdl2cpp/src/converter_complextype.cpp
@@ -285,7 +285,7 @@ QString Converter::generateMemberVariable(const QString &rawName, const QString 
     }
 
     // setter method
-    const QString argName = '_' + memberName;
+    const QString argName = "arg_" + memberName;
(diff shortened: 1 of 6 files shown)


kdwsdl2cpp/src/converter_complextype.cpp
```diff
@@ -389,6 +389,20 @@ QString Converter::generateMemberVariable(const QString &rawName, const QString
     newClass.addFunction(setter);
     newClass.addFunction(getter);
 
+    if (optional) {
+        KODE::Code checkerCode;
+        if (usePointer) {
+            checkerCode += QLatin1String("return ") + variableName + QLatin1String(".isNull() == false;");
```
dfaure-kdab: I would have made it

    return !foo.isNull();

and

    return !foo_nil;

rather than comparing to false, as a more common way of writing this (at least in the Qt world).

PR #164, opened 2019-03-04
List restriction to base

postrstk:
For schema:
```xml
<xsd:simpleType name=dArrayType>
	<xsd:list itemType=xsd:double/>
</xsd:simpleType>

<xsd:simpleType name=d3ArrayType>
	<xsd:restriction base=test:dArrayType>
		<xsd:length value=3/>
	</xsd:restriction>
</xsd:simpleType>
```
when generating code, `kdwsdl2cpp` skips the call to the `entries()` method for `DArrayType`.
For example, `void D3ArrayType::setValue (const DArrayType & value)` instead of `"value.entries().length ()"` is generated by `"value.length()"`
```c++
void D3ArrayType::setValue( const DArrayType& value )
{
    bool rangeOk = true;

    rangeOk = rangeOk && (value.length() == 3);

    if (!rangeOk)
        qDebug( "Invalid range in D3ArrayType::setValue()" );

    mValue = value;
}
```

### DIFF
--- a/kdwsdl2cpp/CMakeLists.txt
+++ b/kdwsdl2cpp/CMakeLists.txt
@@ -5,7 +5,7 @@ include(${QT_USE_FILE})
 remove_definitions(-DQT_NO_CAST_TO_ASCII -DQT_NO_CAST_FROM_ASCII -DQT_NO_CAST_FROM_BYTEARRAY)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
-set(SOURCES
+set(SOURCES_LIB
   common/fileprovider.cpp
   common/messagehandler.cpp
   common/nsmanager.cpp
@@ -59,16 +59,19 @@ set(SOURCES
   src/converter_simpletype.cpp
   src/creator.cpp
   src/elementargumentserializer.cpp
-  src/main.cpp
   src/namemapper.cpp
   src/settings.cpp
   src/typemap.cpp
 )
 
-add_executable(kdwsdl2cpp ${SOURCES})
-target_link_libraries(kdwsdl2cpp ${QT_LIBRARIES})
+set(SOURCES_EXE src/main.cpp)
+
+add_library(kdwsdl2cpp_lib STATIC ${SOURCES_LIB})
+add_executable(kdwsdl2cpp ${SOURCES_EXE})
+
+target_link_libraries(kdwsdl2cpp kdwsdl2cpp_lib ${QT_LIBRARIES})
 
 install(TARGETS kdwsdl2cpp EXPORT "KDSoapTargets"
-  RUNTIME DESTINATION ${INSTALL_RUNTIME_DIR}
-  BUNDLE DESTINATION .
+  DESTINATION ${INSTALL_RUNTIME_DIR}/${PROJECT_NAME_FOR_INSTALL}
+  COMPONENT devel
 )
--- a/kdwsdl2cpp/CMakeLists.txt
+++ b/kdwsdl2cpp/CMakeLists.txt
@@ -72,6 +72,6 @@ add_executable(kdwsdl2cpp ${SOURCES_EXE})
 target_link_libraries(kdwsdl2cpp kdwsdl2cpp_lib ${QT_LIBRARIES})
 
 install(TARGETS kdwsdl2cpp EXPORT "KDSoapTargets"
-  DESTINATION ${INSTALL_RUNTIME_DIR}/${PROJECT_NAME_FOR_INSTALL}
-  COMPONENT devel
+  RUNTIME DESTINATION ${INSTALL_RUNTIME_DIR}
+  BUNDLE DESTINATION .
 )
--- a/kdwsdl2cpp/src/converter_simpletype.cpp
+++ b/kdwsdl2cpp/src/converter_simpletype.cpp
@@ -482,7 +482,7 @@ static KODE::Code createRangeCheckCode(const XSD::SimpleType *type, const QStrin
     QString extendVariableName = variableName;
 
     if (!baseSimpleType.isNull()) {
-        if (baseSimpleType.subType() == 1) {
+        if (baseSimpleType.subType() == XSD::SimpleType::SubType::TypeList) {
             extendVariableName += ".entries()";
         }
         else {
--- a/kdwsdl2cpp/src/converter_simpletype.cpp
+++ b/kdwsdl2cpp/src/converter_simpletype.cpp
@@ -475,8 +475,7 @@ static KODE::Code createRangeCheckCode(const XSD::SimpleType *type, const QStrin
     if (!baseSimpleType.isNull()) {
         if (baseSimpleType.subType() == XSD::SimpleType::SubType::TypeList) {
             extendedVariableName += ".entries()";
-        }
-        else {
+        } else {
             extendedVariableName += ".value()";
         }
     }
--- a/kdwsdl2cpp/src/converter_simpletype.cpp
+++ b/kdwsdl2cpp/src/converter_simpletype.cpp
@@ -229,16 +229,7 @@ void Converter::convertSimpleType(const XSD::SimpleType *type, const XSD::Simple
 
         newClass.addFunction(setter);
         newClass.addFunction(getter);
-<<<<<<< HEAD
-=======
 
-        // getLength
-        //KODE::Function length("length", "int");
-        //length.setBody("return entries().size();");
-        //length.setConst(true);
-
-        //newClass.addFunction(length);
->>>>>>> 8b9c1e6... fix code generate
     }
     break;
     case XSD::SimpleType::TypeUnion:
@@ -479,14 +470,14 @@ static QString escapeRegExp(const QString &str)
 
 static KODE::Code createRangeCheckCode(const XSD::SimpleType *type, const QString &baseTypeName, const QString &variableName, KODE::Class &parentClass, const XSD::SimpleType &baseSimpleType)
 {
-    QString extendVariableName = variableName;
+    QString extendedVariableName = variableName;
 
     if (!baseSimpleType.isNull()) {
         if (baseSimpleType.subType() == XSD::SimpleType::SubType::TypeList) {
-            extendVariableName += ".entries()";
+            extendedVariableName += ".entries()";
         }
         else {
-            extendVariableName += ".value()";
+            extendedVariableName += ".value()";
         }
     }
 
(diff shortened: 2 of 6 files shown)


unittests/unittests.pro
```diff
@@ -37,16 +37,16 @@ SUBDIRS = \
   encapsecurity \
   prefix_wsdl \
   vidyo \
+  list_restriction \
   empty_list_wsdl \
-  onvif_org_event \
```
SudakovIvan: Why onvif_org_event was removed?

unittests/unittests.pro
```diff
@@ -37,16 +37,16 @@ SUBDIRS = \
...
-  empty_element_wsdl \
   fault_namespace \
+  empty_element_wsdl \
   enzo \
   date_example \
   dv_terminalauth \
   test_calc \
   ws_addressing_support \
-  ws_usernametoken_support
```
SudakovIvan: Why was "ws_usernametoken_support" removed?

unittests/unittests.pro
```diff
@@ -37,16 +37,16 @@ SUBDIRS = \
...
+  default_attribute_value_wsdl
```
SudakovIvan: These tests only exist in the Rec-SPb-ETU fork

postrstk: Mistake.
Left tests from KDAB/KDSOAP + "list_restriction"

kdwsdl2cpp/CMakeLists.txt
```diff
@@ -59,16 +59,19 @@ set(SOURCES
...
+
+add_library(kdwsdl2cpp_lib STATIC ${SOURCES_LIB})
+add_executable(kdwsdl2cpp ${SOURCES_EXE})
+
+target_link_libraries(kdwsdl2cpp kdwsdl2cpp_lib ${QT_LIBRARIES})
 
 install(TARGETS kdwsdl2cpp EXPORT "KDSoapTargets"
-  RUNTIME DESTINATION ${INSTALL_RUNTIME_DIR}
-  BUNDLE DESTINATION .
+  DESTINATION ${INSTALL_RUNTIME_DIR}/${PROJECT_NAME_FOR_INSTALL}
```
dfaure-kdab: What is PROJECT_NAME_FOR_INSTALL?

kdwsdl2cpp/CMakeLists.txt
```diff
@@ -59,16 +59,19 @@ set(SOURCES
...
-target_link_libraries(kdwsdl2cpp ${QT_LIBRARIES})
+set(SOURCES_EXE src/main.cpp)
+
+add_library(kdwsdl2cpp_lib STATIC ${SOURCES_LIB})
+add_executable(kdwsdl2cpp ${SOURCES_EXE})
+
+target_link_libraries(kdwsdl2cpp kdwsdl2cpp_lib ${QT_LIBRARIES})
 
 install(TARGETS kdwsdl2cpp EXPORT "KDSoapTargets"
-  RUNTIME DESTINATION ${INSTALL_RUNTIME_DIR}
```
dfaure-kdab: why did you remove RUNTIME?

kdwsdl2cpp/src/converter_simpletype.cpp
```diff
@@ -466,8 +468,19 @@ static QString escapeRegExp(const QString &str)
...
-static KODE::Code createRangeCheckCode(const XSD::SimpleType *type, const QString &baseTypeName, const QString &variableName, KODE::Class &parentClass)
+static KODE::Code createRangeCheckCode(const XSD::SimpleType *type, const QString &baseTypeName, const QString &variableName, KODE::Class &parentClass, const XSD::SimpleType &baseSimpleType)
 {
+    QString extendedVariableName = variableName;
+
+    if (!baseSimpleType.isNull()) {
+        if (baseSimpleType.subType() == XSD::SimpleType::SubType::TypeList) {
+            extendedVariableName += ".entries()";
+        }
+        else {
```
dfaure-kdab: coding style: join with previous line

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+#include <converter.h>
+#include <creator.h>
+#include <settings.h>
+
+#include <compiler.h>
+//#include "wsdl_import_definition.h"
+#include "httpserver_p.h"
+
+#include <QtTest/QtTest>
```
dfaure-kdab: This includes all of QtTest and all of QtCore, please use `QTest` instead

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
...
+#include <QEventLoop>
```
dfaure-kdab: remove, not used

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
+#include <wsdl.h>
+
+#include <fileprovider.h>
```
dfaure-kdab: unused?

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
...
+                                             "</definitions>");
+        QXmlInputSource source;
+        source.setData(srcXml);
+        QXmlSimpleReader reader;
+        reader.setFeature(QLatin1String("http://xml.org/sax/features/namespace-prefixes"), true);
+
+        QString errorMsg;
+        int errorLine, errorCol;
+        QDomDocument doc;
+        doc.setContent(&source, &reader, &errorMsg, &errorLine, &errorCol);
```
dfaure-kdab: wrap in `QVERIFY()`

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
...
+            KWSDL::WSDL wsdl;
+            wsdl.setDefinitions(definitions);
+            wsdl.setNamespaceManager(namespaceManager);
+            KWSDL::Converter converter;
+            converter.setWSDL(wsdl);
+
+            converter.convert();
+            foreach (auto elem, converter.classes()) {
+                if(elem.name() == QString("TEST__D3ArrayType")) {
+                    foreach(auto e_f,elem.functions()) {
```
dfaure-kdab: missing space after comma, and after foreach

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
...
+                        if(e_f.name() == QString("setValue")) {
```
dfaure-kdab: missing space after if

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
...
+            KODE::Code::setDefaultIndentation(4);
+            KWSDL::WSDL wsdl;
+            wsdl.setDefinitions(definitions);
+            wsdl.setNamespaceManager(namespaceManager);
+            KWSDL::Converter converter;
+            converter.setWSDL(wsdl);
+
+            converter.convert();
+            foreach (auto elem, converter.classes()) {
+                if(elem.name() == QString("TEST__D3ArrayType")) {
```
dfaure-kdab: missing space after if

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
...
+        NSManager namespaceManager;
+        namespaceManager.setPrefix(QLatin1String("xml"), NSManager::xmlNamespace());
+
+        MessageHandler messageHandler;
+        ParserContext context;
+        context.setNamespaceManager(&namespaceManager);
+        context.setMessageHandler(&messageHandler);
+
+        KWSDL::Definitions definitions;
+        if (definitions.loadXML(&context, element)) {
```
dfaure-kdab: `QVERIFY()` rather than `if()`

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,358 @@
...
+            definitions.fixUpDefinitions(/*&context, element*/);
+            KODE::Code::setDefaultIndentation(4);
+            KWSDL::WSDL wsdl;
+            wsdl.setDefinitions(definitions);
+            wsdl.setNamespaceManager(namespaceManager);
+            KWSDL::Converter converter;
+            converter.setWSDL(wsdl);
+
+            converter.convert();
+            foreach (auto elem, converter.classes()) {
```
dfaure-kdab: All this test, until this point, is very similar to the previous one.

You could extract a helper function for all three tests, which takes XML as input, and returns a list of classes as output?
With qtestlib that's a bit tricky (QVERIFY/QCOMPARE inside the helper function requires that it returns void), there are several solutions. One solution would be to take `Class::List &` as second argument to the function (and assign to that parameter at the end of the function).

unittests/list_restriction/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+project(list_restriction)
+
+set(list_restriction_SRCS test_list_restriction.cpp)
+set(EXTRA_LIBS kdwsdl2cpp_lib)
```
dfaure-kdab: Unittests for the kdwsdl2cpp code directly, that's interesting. I like the idea, even though in this particular case, the standard way would have been to save those wsdl documents as actual files, running kdwsdl2cpp on them and simply checking that the result compiles, as we do in many other unittesets :-)

But I like it, it can be useful for other bugs that require direct unittests, i.e. those where "it compiles" wouldn't be enough anyway.

kdwsdl2cpp/CMakeLists.txt
postrstk: These are changes from another branch. They are not needed here.

unittests/list_restriction/test_list_restriction.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <parsercontext.h>
+
+#include <converter.h>
+#include <creator.h>
+#include <settings.h>
+
+#include <compiler.h>
+#include "httpserver_p.h"
+
+#include <QtTest>
```
dfaure-kdab: This includes all of QtTest *and* all of QtCore.

I'll change it to <QTest> instead (and add whatever is missing)

dfaure-kdab: (you forgot to remove the 't' so this is still a full-module-include)

PR #173, opened 2019-04-24
rpm / src.rmp build

postrstk:
Added build `rpm` and `src.rpm` packages. Also added the separation of `kdsoap` into `server` `core` and `devel` components.


src/KDSoapServer/CMakeLists.txt
```diff
@@ -73,12 +73,23 @@ if(KDSoap_IS_ROOT_PROJECT)
...
+    DESTINATION ${INSTALL_INCLUDE_DIR}/${PROJECT_NAME_FOR_INSTALL}/KDSoapServer
+    COMPONENT devel
   )
 
   install(TARGETS kdsoap-server EXPORT KDSoapTargets
-    RUNTIME DESTINATION ${INSTALL_RUNTIME_DIR}
-    LIBRARY DESTINATION ${INSTALL_LIBRARY_DIR}
-    ARCHIVE DESTINATION ${INSTALL_ARCHIVE_DIR}
+    RUNTIME DESTINATION ${INSTALL_RUNTIME_DIR}/${PROJECT_NAME_FOR_INSTALL}
+    LIBRARY DESTINATION ${INSTALL_LIBRARY_DIR}/${PROJECT_NAME_FOR_INSTALL}
```
dfaure-kdab: For a prefix of /usr, this will install the library in /usr/lib/KDSoap-qt5/, right? This is very non-standard.

The default prefix is one thing, but installing into /usr should respect the usual directory layout.
Can you look into reverting this part, or letting me know what this is trying to achieve exactly so we can find out how to address that issue better?

SudakovIvan: We need to install kdsoap-qt5 and kdsoap-qt4 simultaneously on the same computer. So, we install them in different folders to avoid conflicts. Actually, there is no need to install libraries in different folders, but the header files should be (because libraries have different names and headers do not have)

dfaure-kdab: OK, a subdir for headers is more acceptable than a subdir for libs, indeed.

However please check that the cmake target points to that include location so that things work out of the box.
This normally requires something like
`target_include_directories(kdsoap INTERFACE "$<INSTALL_INTERFACE:.....>")`

Then whoever does find_package(KDSoap) (which picks up KDSoapConfig.cmake) should be able to just link to KDSoap::kdsoap and the include path will be set correctly.

dfaure-kdab: `target_include_directories(kdsoap INTERFACE "$<INSTALL_INTERFACE:.....>")` has just been added in commit c095f178af092c1e1b490b8ea97ec6a172cd80b1. It just needs to be updated if you install headers elsewhere.

SudakovIvan: Ok, we updated our fork and no need to build src rpm through CMake nowadays, so I'll close this request

SudakovIvan: David or postrstk, please close this

PR #177, opened 2019-07-18
Add c++11 override indicator

caspermeijn:
This adds the c++11 override indicator (1) to the library code, (2) to the generated code and (3) enables warnings for missing indicators.
The library is already dependent on c++11, so commit 1 is not a problem. However I am not sure whether the generated code is already dependant on c++11 and whether this would be a problem (commit 2). The third commit enables a warning in CMake, I am not sure whether this is the correct way to do that.
I want this change as recent compilers will show a lot of warning for the generated code of the ONVIF wsdl.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,10 @@ if(MSVC)
   add_definitions(-D_SCL_SECURE_NO_WARNINGS)
 endif()
 
+if (NOT MSVC)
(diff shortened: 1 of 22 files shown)


dfaure-kdab:
I see no problem with the generated code requiring C++11, since it has to compile on top of the library which already requires C++11 anyway (including in its headers, now).

PR #178, opened 2019-07-19
Add source signature

tuxmaster5000:


winterz:
I do put the .asc files on the github release page so I could change how I do things on the opensuse build system to also include the .asc files.
which means changing my release process and updating my documentation.

but this is the right thing to do for all the KDAB project releases on Github where we provide a spec file
kdreports, kdchart, Gammaray, etc.

PR #180, opened 2019-08-16
fix spec compliance

apachelogger:
as per https://www.w3.org/Submission/ws-addressing/ the only well-known
uri in 2004/08 was anonymous and it has a role prefix. from the
implicit reading I can find on earlier versions the same applies to them.
    
so, when constructing the well-known URIs inject a role prefix when
dealing with specs up to 2004/08 and raise a warning when attempting to
use anything but anonymous

### DIFF
--- a/src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
+++ b/src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
@@ -217,20 +217,41 @@ KDSoapMessageAddressingProperties::~KDSoapMessageAddressingProperties()
 
 QString KDSoapMessageAddressingProperties::predefinedAddressToString(KDSoapMessageAddressingProperties::KDSoapAddressingPredefinedAddress address, KDSoapAddressingNamespace addressingNamespace)
 {
-    const QString addressingNS = addressingNamespaceToString(addressingNamespace);
+    QString prefix = addressingNamespaceToString(addressingNamespace);
+    switch (addressingNamespace) {
+    case Addressing200303:
+    case Addressing200403:
+    case Addressing200408: {
+        switch (address) {
+        case Anonymous:
+            prefix += QLatin1String("/role");
+            break;
+        case Unspecified:
+            prefix += QLatin1String("/id");
+            break;
+        default:
+            qWarning("Anything but Anonymous or Unspecified has no meaning in ws-addressing 2004/08 and earlier");
+            return QString();
+        }
+        break;
+    }
+    default:
+        break;
+    }
+
     switch (address) {
     case Anonymous:
-        return addressingNS + QLatin1String("/anonymous");
+        return prefix + QLatin1String("/anonymous");
     case None:
-        return addressingNS + QLatin1String("/none");
+        return prefix + QLatin1String("/none");
     case Reply:
-        return addressingNS + QLatin1String("/reply");
+        return prefix + QLatin1String("/reply");
     case Unspecified:
-        return addressingNS + QLatin1String("/unspecified");
-    default:
-        Q_ASSERT(false); // should never happen
-        return QString();
+        return prefix + QLatin1String("/unspecified");
     }
+
+    Q_ASSERT(false); // should never happen
+    return QString();
 }
 
 bool KDSoapMessageAddressingProperties::isWSAddressingNamespace(const QString &namespaceUri)
@@ -296,7 +317,19 @@ void KDSoapMessageAddressingProperties::writeMessageAddressingProperties(KDSoapN
     Q_UNUSED(messageNamespace);
     Q_UNUSED(forceQualified);
 
-    if (d->destination == predefinedAddressToString(None, d->addressingNamespace)) {
+    bool supportsNoneAddressing = false;
+    switch (d->addressingNamespace) {
+    case Addressing200303:
+    case Addressing200403:
+    case Addressing200408:
+        supportsNoneAddressing = false;
+        break;
+    case Addressing200508:
+        supportsNoneAddressing = true;
+        break;
+    }
+
+    if (supportsNoneAddressing && d->destination == predefinedAddressToString(None, d->addressingNamespace)) {
         return;
     }
 
--- a/unittests/ws_addressing_support/wsaddressingtest.cpp
+++ b/unittests/ws_addressing_support/wsaddressingtest.cpp
@@ -53,9 +53,15 @@ private Q_SLOTS:
 
         QString unspecified = KDSoapMessageAddressingProperties::predefinedAddressToString(KDSoapMessageAddressingProperties::Unspecified);
         QCOMPARE(unspecified, QString("http://www.w3.org/2005/08/addressing/unspecified"));
-        
-        QString none200303 = KDSoapMessageAddressingProperties::predefinedAddressToString(KDSoapMessageAddressingProperties::None, KDSoapMessageAddressingProperties::Addressing200303);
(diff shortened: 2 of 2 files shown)


src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
```diff
@@ -217,16 +217,34 @@ KDSoapMessageAddressingProperties::~KDSoapMessageAddressingProperties()
...
+    QString prefix = addressingNamespaceToString(addressingNamespace);
+    // Up to and including the 2004/08 spec the well-known URIs have a /role/
+    // folder in their path and furthermore only /anonymous is actually
+    // well-known.
+    switch (addressingNamespace) {
+    case Addressing200303:
+    case Addressing200403:
+    case Addressing200408: {
+        prefix += QLatin1String("/role");
+        if (!address == Anonymous) {
```
dfaure-kdab: Err, did you mean `if (address != Anonymous)` ?

apachelogger: Oh my, yes. Fixed

apachelogger:
@dfaure btw, do you have an opinion WRT adding a Q_ASSERT(false) to the Anonymous condition? At first that looked like a good idea but then testing would be problematic so I opted not to assert anything.

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
```diff
@@ -217,16 +217,34 @@ KDSoapMessageAddressingProperties::~KDSoapMessageAddressingProperties()
...
+    // Up to and including the 2004/08 spec the well-known URIs have a /role/
+    // folder in their path and furthermore only /anonymous is actually
+    // well-known.
+    switch (addressingNamespace) {
+    case Addressing200303:
+    case Addressing200403:
+    case Addressing200408: {
+        prefix += QLatin1String("/role");
+        if (address != Anonymous) {
+            qWarning("Anything but Anonymous has no meaning in ws-addressing 2004/08");
```
caspermeijn: As noted in https://github.com/caspermeijn/KDSoap/pull/1:
I think that a assert would be better. This is more explicit during development. For example:
Q_ASSERT(false); // should never happen

caspermeijn: This warning mentions 2004/08, but is also valid for the other versions.

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
```diff
@@ -217,16 +217,34 @@ KDSoapMessageAddressingProperties::~KDSoapMessageAddressingProperties()
...
+    QString prefix = addressingNamespaceToString(addressingNamespace);
+    // Up to and including the 2004/08 spec the well-known URIs have a /role/
+    // folder in their path and furthermore only /anonymous is actually
+    // well-known.
+    switch (addressingNamespace) {
+    case Addressing200303:
+    case Addressing200403:
+    case Addressing200408: {
+        prefix += QLatin1String("/role");
+        if (address != Anonymous) {
```
caspermeijn: There appears to be a `http://schemas.xmlsoap.org/ws/2004/08/addressing/id/unspecified` in the spec. So I assume this should be added as well.

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
```diff
@@ -217,16 +217,34 @@ KDSoapMessageAddressingProperties::~KDSoapMessageAddressingProperties()
...
+        break;
+    }
+
     switch (address) {
     case Anonymous:
-        return addressingNS + QLatin1String("/anonymous");
+        return prefix + QLatin1String("/anonymous");
     case None:
-        return addressingNS + QLatin1String("/none");
+        return prefix + QLatin1String("/none");
```
caspermeijn: `writeMessageAddressingProperties` uses None. So I assume this will not work properly for the older specs. (If the Q_ASSERT is added, then the unit test should show that :-)

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
apachelogger: asserts now

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
apachelogger: unspecified is now also covered and has a test case

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
apachelogger: writeMessageAddressingProperties now has a guard

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
```diff
@@ -217,20 +217,44 @@ KDSoapMessageAddressingProperties::~KDSoapMessageAddressingProperties()
 
 QString KDSoapMessageAddressingProperties::predefinedAddressToString(KDSoapMessageAddressingProperties::KDSoapAddressingPredefinedAddress address, KDSoapAddressingNamespace addressingNamespace)
 {
-    const QString addressingNS = addressingNamespaceToString(addressingNamespace);
+    QString prefix = addressingNamespaceToString(addressingNamespace);
+    // Up to and including the 2004/08 spec the well-known URIs have a /role/
+    // folder in their path and furthermore only /anonymous is actually
+    // well-known.
```
caspermeijn: This comment doesn't match the code anymore. Maybe the comment is not needed, as it describes the code below.

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
```diff
@@ -296,7 +320,19 @@ void KDSoapMessageAddressingProperties::writeMessageAddressingProperties(KDSoapN
...
+    switch (d->addressingNamespace) {
+    case Addressing200303:
+    case Addressing200403:
+    case Addressing200408:
+        supportsNoneAddressing = false;
+        break;
+    case Addressing200508:
+        supportsNoneAddressing = true;
+        break;
+    }
```
caspermeijn: Would it make sense to just check for Addressing200508 in a if statement?

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
caspermeijn: Great!

dfaure-kdab:
> @dfaure-kdab btw, do you have an opinion WRT adding a Q_ASSERT(false) to the Anonymous condition? At first that looked like a good idea but then testing would be problematic so I opted not to assert anything.

Asserts prevent writing unittests for these conditions. So it might be better to not assert.
(Not to mention the risk that this gets triggered on data outside the control of the developer, such as server replies, or broken XSD/WSDL files)

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
apachelogger: I dislike if statements for enums with a fierce passion. More so in this case because it is in fact >= 200508, or at the very least it needs looking at for possible future versions so the compiler complaining about the incomplete switch-case is much desired here.

apachelogger:
All comments should be dealt with now. I've also switched back to no assert and added back the test case for the bad-value-for-version scenario.

caspermeijn:
Looks good to me

apachelogger:
ping

apachelogger:
actually in #193 for reasons...

PR #181, opened 2019-08-20
Fix rpath for Unix/Linux/macOS

pasnox:


cmake/InstallLocation.cmake
```diff
@@ -26,3 +26,9 @@ endif()
 if(NOT INSTALL_DOC_DIR)
   set(INSTALL_DOC_DIR ${CMAKE_INSTALL_DOCDIR})
 endif()
+
+set(CMAKE_BUILD_WITH_INSTALL_RPATH OFF)
```
dfaure-kdab: Isn't that the default value for this setting anyway?

(I'm phrasing this as a question to be polite, but in fact I'm sure: I double-checked both the documentation and the cmake code, and indeed OFF *is* the default value)

So I would say this line can be removed, it's confusing because it's a no-op.

The rest looks fine to me.

CMakeLists.txt
```diff
@@ -131,9 +131,6 @@ if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
   set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/lib")
   set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/lib")
 
-  set(CMAKE_INSTALL_RPATH ${INSTALL_LIBRARY_DIR})
```
dfaure-kdab: Oops, this line should have been set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})

Right now it adds "lib64" to the RPATH instead of the full path, and as a result I get

    $ ldd /d/kdab/inst/KDSoap-Qt5-cmake/lib64/libkdsoap-server.so.1.8.50
            libkdsoap.so.1.8.50 => not found

But your patch still has that problem, missing RPATH for kdsoap-server.so to find kdsoap.so in the same install dir.
If I change this line to set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR}) then it works.
Make install says

    Set runtime path of "/d/kdab/inst/KDSoap-Qt5-cmake/lib64/libkdsoap-server.so.1.8.50" to "/d/kdab/inst/KDSoap-Qt5-cmake/lib64:/d/qt/5/inst/lib"

and ldd says

    libkdsoap.so.1.8.50 => /d/kdab/inst/KDSoap-Qt5-cmake/lib64/libkdsoap.so.1.8.50 (0x00007f2bbdf20000)

Which problem does this setting create for you?

cmake/InstallLocation.cmake
pasnox: That's indeed already the default, I suppose it was just artifacts from my old code i copied from.

CMakeLists.txt
pasnox: My apologies, i did not properly tested the linux variant :/
My new changes properly set $ORIGIN for non apple platforms.
The original issues was non out of the box works on mac os when linked from its build folder (but it may also be a kdabviewer external add project issue) and non proper rpath for installed libraries on mac os.
Should now be fine on all platforms and works correctly in regards to (relative) RPATH.

PR #185, opened 2019-08-23
initialize with nullptr, since C++11 is already required

### LINKED ISSUE #182: kdwsdl2cpp generates 0 instand of nullptr
The code generated by it will have lines like this:

> QObject* _parent = 0

But to support the modern features of the compilers, it will be better to use nullptr instant of 0.
This change will increase the code security, because nullptr is an part of C++11

anoymouserver:
fixes #182

### DIFF
--- a/kdwsdl2cpp/common/fileprovider.cpp
+++ b/kdwsdl2cpp/common/fileprovider.cpp
@@ -41,7 +41,7 @@ static QHash<QUrl, QByteArray> fileProviderCache;
 
 
 FileProvider::FileProvider()
-  : QObject( 0 )
+  : QObject( nullptr )
 {
 }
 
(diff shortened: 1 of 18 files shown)


kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -41,7 +41,7 @@ static QHash<QUrl, QByteArray> fileProviderCache;
 
 
 FileProvider::FileProvider()
-  : QObject( 0 )
+  : QObject( nullptr )
```
ivan-cukic: This constructor can be `= default` or removed altogether.

anoymouserver: I know, but I didn't wanted to change the basic layout, otherwise there are also some destructors which could be replaced.

PR #186, opened 2019-08-24
KDSoapUdpClient

### LINKED ISSUE #175: SOAP over UDP
Hi, I am building a [library](https://gitlab.com/caspermeijn/kdsoap-ws-discovery-client) to do [WS-Discovery](http://docs.oasis-open.org/ws-dd/discovery/1.1/os/wsdd-discovery-1.1-spec-os.html). For this I need to support sending and receiving [SOAP over UDP](http://docs.oasis-open.org/ws-dd/soapoverudp/1.1/os/wsdd-soapoverudp-1.1-spec-os.html). I got this working by including KDSoapMessageReader_p.h and KDSoapMessageWriter_p.h for the KDSoap sources. These classes are exported for the KDSoapServer library, but are not part of the official api.

I see three ways to make this work properly:
1. Make the classes part of the public api.
2. Add a QUdpSocket wrapper, which can send and receive KDSoapMessages (something like [this](https://gitlab.com/caspermeijn/kdsoap-ws-discovery-client/blob/master/src/soapudpclient.h)).
3. Create a new [Binding type](https://github.com/KDAB/KDSoap/blob/master/kdwsdl2cpp/wsdl/binding.h#L45) and let kdwsdl2cpp generate a UDP services.

I think it is unwanted to export the internal apis KDSoapMessageReader and KDSoapMessageWriter. And I think 3 is a bit overkill as I think WS-Discovery will be the only user of UDP. I think 2 is the best solution.

What do you think about this? Would you accept something like [SoapUdpClient](https://gitlab.com/caspermeijn/kdsoap-ws-discovery-client/blob/master/src/soapudpclient.h)? Or would you prefer a different solution to this problem?

caspermeijn:
KDSoapUdpClient provides an interface for implementing a
SOAP-over-UDP[1] client.

It includes tests and documentation. I tested it with (kdsoap-ws-discovery-client)[https://gitlab.com/caspermeijn/kdsoap-ws-discovery-client].

Fixes #175 

[1] https://docs.oasis-open.org/ws-dd/soapoverudp/1.1/os/wsdd-soapoverudp-1.1-spec-os.html

### DIFF
--- a/src/KDSoapClient/KDSoapUdpClient.cpp
+++ b/src/KDSoapClient/KDSoapUdpClient.cpp
@@ -1,0 +1,134 @@
+/*
+ * Copyright (C) 2019  Casper Meijn <casper@meijn.net>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "KDSoapUdpClient.h"
+#include "KDSoapUdpClient_p.h"
+
+#include "KDSoapMessage.h"
+#include "KDSoapMessageReader_p.h"
+#include "KDSoapMessageWriter_p.h"
+#include <QNetworkInterface>
+
+static const QHostAddress AnyIPv4(QLatin1String("0.0.0.0"));
+
+static bool isMulticastAddress(const QHostAddress &address) {
+    if (address.protocol() == QAbstractSocket::IPv4Protocol) {
+        return address.isInSubnet(QHostAddress (QLatin1String("224.0.0.0")), 4);
+    } else if (address.protocol() == QAbstractSocket::IPv6Protocol) {
+        return address.isInSubnet(QHostAddress (QLatin1String("ff00::")), 8);
+    }
+    return false;
+}
+
+KDSoapUdpClient::KDSoapUdpClient(QObject *parent) : 
+    QObject(parent),
+    d_ptr(new KDSoapUdpClientPrivate(this))
+{
+    Q_D(KDSoapUdpClient);
+    d->socketIPv4 = new QUdpSocket(this);
+    connect(d->socketIPv4, SIGNAL(readyRead()), d, SLOT(readyRead()));
+    d->socketIPv6 = new QUdpSocket(this);
+    connect(d->socketIPv6, SIGNAL(readyRead()), d, SLOT(readyRead()));
+}
+
+KDSoapUdpClient::~KDSoapUdpClient()
+{
+    delete d_ptr;
+}
+
+bool KDSoapUdpClient::bind(quint16 port, QAbstractSocket::BindMode mode) {
+    Q_D(KDSoapUdpClient);
+    bool rc = true;
+    // Workaround for lack of dual stack sockets in Qt4
+    // Qt5 supports binding to QHostAddress::Any, which will listen on both IPv4 and IPv6 interfaces.
+    rc = d->socketIPv4->bind(AnyIPv4, port, mode) && rc;
+    rc = d->socketIPv6->bind(QHostAddress::AnyIPv6, port, mode) && rc;
+    return rc;
+}
+
+void KDSoapUdpClient::setSoapVersion(KDSoap::SoapVersion version) {
+    Q_D(KDSoapUdpClient);
+    d->soapVersion = version;
+}
+
+bool KDSoapUdpClient::sendMessage(const KDSoapMessage &message, const KDSoapHeaders &headers, const QHostAddress &address, quint16 port) {
+    Q_D(KDSoapUdpClient);
+    KDSoapMessageWriter msgWriter;
+    msgWriter.setVersion(d->soapVersion);
+    const QByteArray data = msgWriter.messageToXml(message, QString(), headers, QMap<QString, KDSoapMessage>());
+    
+    if (isMulticastAddress(address)) {
+        bool anySuccess = false;
+        const auto& allInterfaces = QNetworkInterface::allInterfaces();
+        for (const auto &iface : allInterfaces) {
+            if (iface.flags().testFlag(QNetworkInterface::IsUp) &&
+                iface.flags().testFlag(QNetworkInterface::CanMulticast)) {
+                //qDebug() << "Sending multicast to" << iface.name() << address << ":" << data;
+                if (address.protocol() == QAbstractSocket::IPv4Protocol) {
+                    d->socketIPv4->setMulticastInterface(iface);
+                    qint64 writtenSize = d->socketIPv4->writeDatagram(data, address, port);
+                    anySuccess = anySuccess || (writtenSize == data.size());
+                } else if (address.protocol() == QAbstractSocket::IPv6Protocol) {
+                    d->socketIPv6->setMulticastInterface(iface);
+                    qint64 writtenSize = d->socketIPv6->writeDatagram(data, address, port);
+                    anySuccess = anySuccess || (writtenSize == data.size());
+                }
+            }
+        }
+        return anySuccess;
+    } else {
+        //qDebug() << "Sending to" << address << ":" << data;
+        if (address.protocol() == QAbstractSocket::IPv4Protocol) {
+            qint64 writtenSize = d->socketIPv4->writeDatagram(data, address, port);
+            return writtenSize == data.size();
+        } else if (address.protocol() == QAbstractSocket::IPv6Protocol) {
+            qint64 writtenSize = d->socketIPv6->writeDatagram(data, address, port);
+            return writtenSize == data.size();
+        }
+    }
+    return false;
+}
+
+void KDSoapUdpClientPrivate::readyRead()
+{
+    QUdpSocket *socket = qobject_cast<QUdpSocket *>(sender());
+    while (socket->hasPendingDatagrams()) {
+        qint64 size = socket->pendingDatagramSize();
+        
+        QByteArray buffer;
+        buffer.resize(size);
+        QHostAddress senderAddress;
+        quint16 senderPort;
+        socket->readDatagram(buffer.data(), buffer.size(), &senderAddress, &senderPort);
+        
+        receivedDatagram(buffer, senderAddress, senderPort);
+    }
+}
+
+void KDSoapUdpClientPrivate::receivedDatagram(const QByteArray &messageData, const QHostAddress &senderAddress, quint16 senderPort)
+{
+    Q_Q(KDSoapUdpClient);
+    //qDebug() << "Received datagram from:" << senderAddress << "data:" << QString::fromUtf8(messageData);
+
+    KDSoapMessage replyMessage;
+    KDSoapHeaders replyHeaders;
+
+    KDSoapMessageReader reader;
+    reader.xmlToMessage(messageData, &replyMessage, 0, &replyHeaders, soapVersion);
+
+    emit q->receivedMessage(replyMessage, replyHeaders, senderAddress, senderPort);
+}
--- a/src/KDSoapClient/KDSoapUdpClient.h
+++ b/src/KDSoapClient/KDSoapUdpClient.h
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2019  Casper Meijn <casper@meijn.net>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef KDSOAPUDPCLIENT_H
+#define KDSOAPUDPCLIENT_H
+
+#include "KDSoapGlobal.h"
+#include "KDSoapClientInterface.h"
+
+#include <QAbstractSocket>
+#include <QObject>
+
+class KDSoapHeaders;
+class KDSoapMessage;
+class KDSoapUdpClientPrivate;
+QT_BEGIN_NAMESPACE
+class QHostAddress;
+QT_END_NAMESPACE
+
+/**
+ * \brief KDSoapUdpClient provides an interface for implementing a [SOAP-over-UDP](https://docs.oasis-open.org/ws-dd/soapoverudp/1.1/os/wsdd-soapoverudp-1.1-spec-os.html) client.
+ * 
+ * One-way SOAP-over-UDP can be send by simply using sendMessage().
+ * 
+ * Request-response SOAP-over-UDP is supported by bind()ing to a sender UDP 
+ * port.  You can send the request using sendMessage() and the response will is
+ * signaled using receivedMessage(). receivedMessage() will signal any response, 
+ * including those of other requests, there is no help with finding the correct 
+ * response. The WS-Addressing properties of the message can be used to filter 
+ * the received responses.
+ * 
+ * \code
+ * auto soapUdpClient = new KDSoapUdpClient(this);
+ * connect(soapUdpClient, &KDSoapUdpClient::receivedMessage, [=](const KDSoapMessage &message, const KDSoapHeaders &headers, const QHostAddress &address, quint16 port) { 
+ *   if(message.messageAddressingProperties().action() == QStringLiteral("http://docs.oasis-open.org/ws-dd/ns/discovery/2009/01/ProbeMatches")) {
+ *     TNS__ProbeMatchesType probeMatches;
+ *     probeMatches.deserialize(message);
+ *     qDebug() << "Received probe match from" << address;
+ *   }
+ * });
+ * soapUdpClient->bind(3702);
+ * 
+ * TNS__ProbeType probe;
+ *
+ * KDSoapMessage message;
+ * message = probe.serialize(QStringLiteral("Probe"));
+ * message.setUse(KDSoapMessage::LiteralUse);
+ * message.setNamespaceUri(QStringLiteral("http://docs.oasis-open.org/ws-dd/ns/discovery/2009/01"));
+ *
+ * KDSoapMessageAddressingProperties addressing;
+ * addressing.setAction(QStringLiteral("http://docs.oasis-open.org/ws-dd/ns/discovery/2009/01/Probe"));
+ * addressing.setMessageID(QStringLiteral("urn:uuid:") + QUuid::createUuid().toString(QUuid::WithoutBraces));
+ * addressing.setDestination(QStringLiteral("urn:docs-oasis-open-org:ws-dd:ns:discovery:2009:01"));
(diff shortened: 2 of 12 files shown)


caspermeijn:
@apachelogger Can you have a look at this?

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,111 @@
...
+            return writtenSize == data.size();
+        }
+    }
+    return false;
+}
+
+void KDSoapUdpClientPrivate::readyReadIPv4()
+{
+    while (socketIPv4->hasPendingDatagrams()) {
+        QNetworkDatagram datagram = socketIPv4->receiveDatagram();
```
apachelogger: The two lines could be merged.
In fact, considering this and ipv6 are the same code except for the socket you could just pass the socket in, or cast sender().

src/KDSoapClient/KDSoapUdpClient.h
```diff
@@ -0,0 +1,126 @@
...
+     * \param message The parsed message received over the socket. Use 
+     * KDSoapMessage::messageAddressingProperties() to see who the recipient is 
+     * and what SOAP action is requested.
+     * \param headers The additional headers of the message
+     * \param address The IP-address of the sender
+     * \param port The UDP port of the sender
+     * \see bind()
+     * \since 1.9
+     */
+    void receivedMessage(const KDSoapMessage& message, const KDSoapHeaders& headers, const QHostAddress& address, quint16 port);
```
apachelogger: & on wrong side of space

src/KDSoapClient/KDSoapUdpClient.h
```diff
@@ -0,0 +1,126 @@
...
+     * \param message  The actual message to be send. Use 
+     * KDSoapMessage::setMessageAddressingProperties() to set the WS-Addressing 
+     * properties of the message.
+     * \param headers can be used to add additional SOAP headers.
+     * \param address The address to send to message to. Messages send to a 
+     * multicast address will be send to to all interfaces.
+     * \param port The UDP port to send the message to
+     * \since 1.9
+     */    
+    bool sendMessage(const KDSoapMessage &message, const KDSoapHeaders &headers, const QHostAddress& address, quint16 port);
```
apachelogger: & on wrong side of space

src/KDSoapClient/KDSoapUdpClient_p.h
```diff
@@ -0,0 +1,41 @@
...
+#ifndef KDSOAPUDPCLIENT_P_H
+#define KDSOAPUDPCLIENT_P_H
+
+#include <QObject>
+#include <QUdpSocket>
+
+class KDSoapUdpClientPrivate
+{
+public:
+    KDSoapUdpClientPrivate(KDSoapUdpClient * q) : q_ptr(q){;}
```
apachelogger: unnecessary ;

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,111 @@
...
+    d->socketIPv4 = new QUdpSocket(this);
+    connect(d->socketIPv4, &QUdpSocket::readyRead, [d]() { d->readyReadIPv4(); });
+    d->socketIPv6 = new QUdpSocket(this);
+    connect(d->socketIPv6, &QUdpSocket::readyRead, [d]() { d->readyReadIPv6(); });
+}
+
+bool KDSoapUdpClient::bind(quint16 port) {
+    Q_D(KDSoapUdpClient);
+    bool rc = true;
+    rc = d->socketIPv4->bind(QHostAddress::AnyIPv4, port, QUdpSocket::ShareAddress) && rc;
```
apachelogger: Is there a particular reason the ports need to be shared?

dfaure-kdab:
>     * I think kdsoap is still meant to build with Qt4, so you cannot use the lambda based connect syntax. Oh and I think you need qt includes with their directory so <QtCore/QObject> etc.

No, <QObject> works with both Qt4 and Qt5.

>     * I would also use `&=` and `||=` instead of `= foo & bar` respectively `= foo || bar`. The former are distinctly more readable IMHO.

`||=` does not exist in C++, and `&=` would do the wrong thing for boolean expressions. Please stick to `foo = foo && bar` and `foo = foo || bar`, in an order that matches whether you want short circuit evaluation or not.

caspermeijn:
>     * I think kdsoap is still meant to build with Qt4, so you cannot use the lambda based connect syntax. Oh and I think you need qt includes with their directory so <QtCore/QObject> etc.

I made the code compatible with Qt4. However I could only compile test it.

> 
>     * Still some disabled qdebugs littered about.

This appears the way to add optional debug messages to KDSoap.

> 
>     * I would also use `&=` and `||=` instead of `= foo & bar` respectively `= foo || bar`. The former are distinctly more readable IMHO.
> 
>     * For the tests I'd bind port=0 to dynamically get a free port.

I was able to do this for the sending test. However I have to add extra code to make is happens receiving code. Do you think that is worth the trouble?

> 
>     * I am not sure how consistent kdsoap's coding style is but even within the diff the use of spaces is inconsistent. I'd suggest putting a space after every if `if (`, as well as a space before * and & following a type `const QHostAddress &`.

I think I fixed most of those.

> 
> 
> On the not nit-picky front: Since this client is meant to enable ws-discovery implementations, the current code doesn't look quite sufficient for [ws-discovery](http://specs.xmlsoap.org/ws/2005/04/discovery/ws-discovery.pdf)'s needs. Specifically:
> 
> > Messages sent over UDP MUST be sent using SOAP over UDP [SOAP/UDP]. To compensate for possible UDP unreliability, senders MUST use the example transmission algorithm in Appendix I of SOAP over UDP.
> 
> Now, while discovery built on top of the client could do that, I think it stands to reason that it's more expected and efficient if we do it in the client. In particular since all request-response uses of this client will suffer from the inherit unreliability of UDP, so most uses need retransmission to overcome UDP's limitations.

This code doesn't implement a full request-response client, as it doesn't match responses to request. Therefore it can't know when to send a retransmission. I think the request-response should be implemented in another layer above KDSoapUdpClient.

> 
> Other than that itooks lovely.

I added my changes in separate commits for review-ability. I will squash the commits later.

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "KDSoapMessage.h"
+#include "KDSoapMessageReader_p.h"
+#include "KDSoapMessageWriter_p.h"
+#include <QNetworkDatagram>
+#include <QNetworkInterface>
+
+KDSoapUdpClient::KDSoapUdpClient(QObject *parent) : 
+    QObject(parent),
+    d_ptr(new KDSoapUdpClientPrivate(this))
```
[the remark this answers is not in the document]
caspermeijn: Fixed

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,111 @@
...
+{
+    Q_D(KDSoapUdpClient);
+    d->socketIPv4 = new QUdpSocket(this);
```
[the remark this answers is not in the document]
caspermeijn: I believe it isn't leaking, as the socket has a parent.

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,111 @@
...
+            if(iface.flags().testFlag(QNetworkInterface::IsUp) &&
+                iface.flags().testFlag(QNetworkInterface::CanMulticast)) {
+                //qDebug() << "Sending multicast to" << iface.name() << address << ":" << data;
+                if(address.protocol() == QAbstractSocket::IPv4Protocol) {
+                    d->socketIPv4->setMulticastInterface(iface);
+                    qint64 writtenSize = d->socketIPv4->writeDatagram(QNetworkDatagram(data, address, port));
+                    anySuccess = (writtenSize == data.size()) || anySuccess;
+                } else if(address.protocol() == QAbstractSocket::IPv6Protocol) {
+                    d->socketIPv6->setMulticastInterface(iface);
+                    qint64 writtenSize = d->socketIPv6->writeDatagram(QNetworkDatagram(data, address, port));
```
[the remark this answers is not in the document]
caspermeijn: The Qt documentation is pretty sparse. However, it worked on my laptop...

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,111 @@
...
+    bool rc = true;
+    rc = d->socketIPv4->bind(QHostAddress::AnyIPv4, port, QUdpSocket::ShareAddress) && rc;
+    rc = d->socketIPv6->bind(QHostAddress::AnyIPv6, port, QUdpSocket::ShareAddress) && rc;
+    return rc;
+}
+
+bool KDSoapUdpClient::sendMessage(const KDSoapMessage &message, const KDSoapHeaders &headers, const QHostAddress& address, quint16 port) {
+    Q_D(KDSoapUdpClient);
+    KDSoapMessageWriter msgWriter;
+    msgWriter.setVersion(KDSoap::SOAP1_2);
```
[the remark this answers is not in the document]
caspermeijn: Added for sending messages. However this is more difficult for receiving. Should I just assume that the response will be in the same SOAP version as the request or do I need to detect the SOAP version?

src/KDSoapClient/KDSoapUdpClient.cpp
caspermeijn: I changed that code, also for Qt4 compatibility.

src/KDSoapClient/KDSoapUdpClient.cpp
caspermeijn: I was under the impression that other program could use the port as well. This would be useful if multiple programs want to do WS-Discovery.

src/KDSoapClient/KDSoapUdpClient.cpp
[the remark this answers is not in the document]
apachelogger: Mh, I tried my three million interfaces and it works indeed :ok_hand:

src/KDSoapClient/KDSoapUdpClient.cpp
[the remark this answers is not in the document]
apachelogger: I honestly don't know.
KDSoapPendingCall however seems to assume that the reply will be the same version, which does make sense to me anyway.

src/KDSoapClient/KDSoapUdpClient.cpp
apachelogger: Only for the server though. Clients have arbitrary ports, no? E.g. I see windows10 send Probes from all sorts of ports, but the destination is always 3702.

apachelogger:
> I was able to do this for the sending test. However I have to add extra code to make is happens receiving code. Do you think that is worth the trouble?

Don't think so, should it become a problem in the future one can always put in the extra work.

src/KDSoapClient/KDSoapUdpClient.cpp
caspermeijn: You are right, this appear to work, I will allow both uses then. For some reason my firewall will refuse responses to the random port :-(

apachelogger:
looks good to me :+1:

apachelogger:
ping (:

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,133 @@
...
+        socket->readDatagram(buffer.data(), buffer.size(), &senderAddress, &senderPort);
+        
+        recievedDatagram(buffer, senderAddress, senderPort);
+    }
+}
+
+void KDSoapUdpClientPrivate::recievedDatagram(const QByteArray &messageData, const QHostAddress &senderAddress, quint16 senderPort)
+{
+    Q_Q(KDSoapUdpClient);
+    //qDebug() << "Recieved datagram from:" << senderAddress << "data:" << QString::fromUtf8(messageData);
```
dfaure-kdab: typo: Received

unittests/soap_over_udp/test_soap_over_udp.cpp
```diff
@@ -0,0 +1,173 @@
...
+        QSignalSpy spy(&udpClient, SIGNAL(receivedMessage(KDSoapMessage, KDSoapHeaders, QHostAddress, quint16)));
+        
+        auto data = exampleTextData();
+        qint64 size = testSocket.writeDatagram(data, QHostAddress::LocalHost, 12345);
+        QVERIFY(size == data.size());
+        
+        QVERIFY(spy.wait(1000));
+        QList<QVariant> arguments = spy.takeFirst();
+        QVERIFY(KDSoapUnitTestHelpers::xmlBufferCompare(arguments.at(0).value<KDSoapMessage>().toXml(), exampleMessage().toXml()));
+        QVERIFY(arguments.at(1).value<KDSoapHeaders>().size() == 0);
```
dfaure-kdab: QCOMPARE

unittests/soap_over_udp/test_soap_over_udp.cpp
```diff
@@ -0,0 +1,173 @@
...
+        addressing.setReplyEndpointAddress(KDSoapMessageAddressingProperties::predefinedAddressToString(KDSoapMessageAddressingProperties::Anonymous));
+        message.setMessageAddressingProperties(addressing);
+
+        soapUdpClient->sendMessage(message, KDSoapHeaders(), QHostAddress("239.255.255.250"), 3702);
+    }
+    
+    void testSendMessage() {
+        QUdpSocket testSocket;
+        bool rc = testSocket.bind();
+        Q_ASSERT(rc);
```
dfaure-kdab: QVERIFY

unittests/soap_over_udp/test_soap_over_udp.cpp
```diff
@@ -0,0 +1,173 @@
...
+        
+        KDSoapUdpClient udpClient;
+        bool rc = udpClient.bind(12345);
+        QVERIFY(rc);
+        
+        QSignalSpy spy(&udpClient, SIGNAL(receivedMessage(KDSoapMessage, KDSoapHeaders, QHostAddress, quint16)));
+        
+        auto data = exampleTextData();
+        qint64 size = testSocket.writeDatagram(data, QHostAddress::LocalHost, 12345);
+        QVERIFY(size == data.size());
```
dfaure-kdab: QCOMPARE (so that we can see values in case of failure)

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include "KDSoapUdpClient_p.h"
+
+#include "KDSoapMessage.h"
+#include "KDSoapMessageReader_p.h"
+#include "KDSoapMessageWriter_p.h"
+#include <QNetworkInterface>
+
+static const QHostAddress AnyIPv4(QLatin1String("0.0.0.0"));
+
+bool isMulticastAddress(const QHostAddress &address) {
```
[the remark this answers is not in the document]
caspermeijn: Good point

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+bool KDSoapUdpClient::sendMessage(const KDSoapMessage &message, const KDSoapHeaders &headers, const QHostAddress &address, quint16 port) {
+    Q_D(KDSoapUdpClient);
+    KDSoapMessageWriter msgWriter;
+    msgWriter.setVersion(d->soapVersion);
+    const QByteArray data = msgWriter.messageToXml(message, QString(), headers, QMap<QString, KDSoapMessage>());
+    
+    if (isMulticastAddress(address)) {
+        bool anySuccess = false;
+        for (const auto &iface : QNetworkInterface::allInterfaces()) {
```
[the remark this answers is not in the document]
caspermeijn: Done

src/KDSoapClient/KDSoapUdpClient.cpp
caspermeijn: Fixed, also in function name

unittests/soap_over_udp/test_soap_over_udp.cpp
caspermeijn: Makes sense

src/KDSoapClient/KDSoapUdpClient_p.h
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef KDSOAPUDPCLIENT_P_H
+#define KDSOAPUDPCLIENT_P_H
+
+#include <QObject>
+#include <QUdpSocket>
+
```
anoymouserver: I had to include `KDSoapValue.h` (used for `KDSoap::SoapVersion` in line 40) and forward-declare `KDSoapUdpClient` to build it.

apachelogger:
qmake support in #192

src/KDSoapClient/KDSoapUdpClient.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+    delete d_ptr;
+}
+
+bool KDSoapUdpClient::bind(quint16 port, QAbstractSocket::BindMode mode) {
+    Q_D(KDSoapUdpClient);
+    bool rc = true;
+    // Workaround for lack of dual stack sockets in Qt4
+    // Qt5 supports binding to QHostAddress::Any, which will listen on both IPv4 and IPv6 interfaces.
+    rc = d->socketIPv4->bind(AnyIPv4, port, mode) && rc;
```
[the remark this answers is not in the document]
caspermeijn: I am surprised that the code was not compatible with Qt4. I did build this patch using Qt4 to make sure it works. Maybe I build an earlier version and broke the build after that.

In your commit fixing the Qt4 build you mention "once we drop Qt4 support". Are there any plans for dropping Qt4 support? In what time frame could that be expected? Would it be acceptable to make this feature Qt5 only? 

The recent PKCS12 feature only works with Qt5.4, because of the API improvements in that version. But this feature should be able to work correctly using Qt4.

Personally I don't use Qt4 and I can't easily test using Qt4 either (as you have noticed by me breaking Qt4 builds several times ;-). I would happily modify KDSoapUdpClient to be Qt5 only and remove the workaround.

unittests/soap_over_udp/test_soap_over_udp.cpp
```diff
@@ -0,0 +1,173 @@
...
+    void testSendMessage() {
+        QUdpSocket testSocket;
+        bool rc = testSocket.bind();
+        QVERIFY(rc);
+        
+        KDSoapUdpClient udpClient;
+        
+        udpClient.sendMessage(exampleMessage(), KDSoapHeaders(), QHostAddress::LocalHost, testSocket.localPort());
+        
+        QVERIFY(testSocket.hasPendingDatagrams());
```
dfaure-kdab: This test just failed on OSX:

********* Start testing of SoapOverUdpTest *********
Config: Using QtTest library 5.8.0, Qt 5.8.0 (x86_64-little_endian-lp64 shared (dynamic) release build; by Clang 7.0.2 (clang-700.1.81) (Apple))
PASS   : SoapOverUdpTest::initTestCase()
PASS   : SoapOverUdpTest::testExample()
FAIL!  : SoapOverUdpTest::testSendMessage() 'testSocket.hasPendingDatagrams()' returned FALSE. ()
   Loc: [/Users/build/Buildbot/kdsoap-qt5.8-osx2/source/unittests/soap_over_udp/test_soap_over_udp.cpp(94)]
PASS   : SoapOverUdpTest::testReceiveMessage()
PASS   : SoapOverUdpTest::cleanupTestCase()
Totals: 4 passed, 1 failed, 0 skipped, 0 blacklisted, 5ms
********* Finished testing of SoapOverUdpTest *********

Maybe this should use QTRY_VERIFY? I can see how the OS does not guarantee that it's immediately available.

caspermeijn: Sounds reasonable. This was one of the reasons that I added to verify, to ensure the message has arrived before proceeding.

caspermeijn: Do you want me to create this change and the change for disabling KDSoapUdpClient for Qt4?

dfaure-kdab: I'll test the change and I'll come back to you if it still fails.

The unittest is already disabled for Qt4, and we're dropping Qt4 support after the upcoming release, so that's fine.

PR #187, opened 2019-09-02
[kdwsdl2cpp]Add option for loading PKCS12 certificates

martonmiklos:
I needed to interact with a web service protected with PKCS12 based certificate. 
To parse the WSDL files from these servers I added two command line arguments to the kdwsdl2cpp to being able to load a certificate file. 

Ps. because the QSslCertificate::importPkcs12 is only supported above Qt 5.4 and newer I have added guards around the new functionality.

### DIFF
--- a/kdwsdl2cpp/common/fileprovider.cpp
+++ b/kdwsdl2cpp/common/fileprovider.cpp
@@ -106,6 +106,15 @@ bool FileProvider::get( const QUrl &url, QString &target )
 
     QNetworkAccessManager manager;
     QNetworkRequest request(url);
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+    if (Settings::self()->certificateLoaded()) {
+        QSslConfiguration sslConfig = request.sslConfiguration();
+        sslConfig.setPrivateKey(Settings::self()->sslKey());
+        sslConfig.setLocalCertificate(Settings::self()->certificate());
+        sslConfig.setCaCertificates(Settings::self()->caCertificates());
+        request.setSslConfiguration(sslConfig);
+    }
+#endif
     QNetworkReply* job = manager.get(request);
 
     QEventLoop loop;
--- a/kdwsdl2cpp/src/main.cpp
+++ b/kdwsdl2cpp/src/main.cpp
@@ -67,6 +67,12 @@ static void showHelp(const char *appName)
             "                            use of the import-path option\n"
             "  -help-on-missing          When groups or basic types could not be found, display\n"
             "                            available types (helps with wrong namespaces)\n"
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+            "  -pkcs12file               Load a certificate from a PKCS12 file. You can use this option\n"
+            "                            if the WSDL file (or files refering to it) is served from a \n"
+            "                            location which require certificate based authentication\n"
+            "  -pkcs12password           Pass the password for the certificate file if required\n"
+#endif
             "\n", appName, appName, appName);
 }
 
@@ -90,6 +96,7 @@ int main(int argc, char **argv)
     QStringList importPathList;
     bool useLocalFilesOnly = false;
     bool helpOnMissing = false;
+    QString pkcs12File, pkcs12Password;
 
     int arg = 1;
     while (arg < argc) {
@@ -207,6 +214,22 @@ int main(int argc, char **argv)
             useLocalFilesOnly = true;
         } else if (opt == QLatin1String("-help-on-missing")) {
             helpOnMissing = true;
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+        } else if (opt == QLatin1String("-pkcs12file")) {
+            ++arg;
+            if (!argv[arg]) {
+                showHelp(argv[0]);
+                return 1;
+            }
+            pkcs12File = QLatin1String(argv[arg]);
+        } else if (opt == QLatin1String("-pkcs12password")) {
+            ++arg;
+            if (!argv[arg]) {
+                showHelp(argv[0]);
+                return 1;
+            }
+            pkcs12Password = QLatin1String(argv[arg]);
+#endif
         } else if (!fileName) {
             fileName = argv[arg];
         } else {
@@ -262,6 +285,12 @@ int main(int argc, char **argv)
     Settings::self()->setHelpOnMissing(helpOnMissing);
 
     KWSDL::Compiler compiler;
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+    if (!pkcs12File.isEmpty()) {
+        if (!Settings::self()->loadCertificate(pkcs12File, pkcs12Password))
+            return -1;
+    }
+#endif
 
     // so that we have an event loop, for downloads
     QTimer::singleShot(0, &compiler, SLOT(run()));
--- a/kdwsdl2cpp/src/settings.cpp
+++ b/kdwsdl2cpp/src/settings.cpp
@@ -251,3 +251,51 @@ bool Settings::helpOnMissing() const
 {
     return mHelpOnMissing;
 }
+
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+bool Settings::loadCertificate(const QString &certPath, const QString & password)
+{
+    QFile certFile(certPath);
+    if (certFile.open(QFile::ReadOnly)) {
+        mCertificateLoaded = QSslCertificate::importPkcs12(&certFile,
+                                                     &mSslKey,
+                                                     &mCertificate,
+                                                     &mCaCertificates,
+                                                     password.toLocal8Bit());
+        certFile.close();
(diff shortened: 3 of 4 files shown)


kdwsdl2cpp/src/main.cpp
```diff
@@ -67,6 +68,12 @@ static void showHelp(const char *appName)
             "                            use of the import-path option\n"
             "  -help-on-missing          When groups or basic types could not be found, display\n"
             "                            available types (helps with wrong namespaces)\n"
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
```
caspermeijn: I don't like it that the option is dependent on de Qt version. However I don't know how to do it better ;-)

kdwsdl2cpp/common/fileprovider.h
```diff
@@ -40,8 +43,18 @@ class KXMLCOMMON_EXPORT FileProvider : QObject
...
+    static bool loadCertificate(const QString & certPath, const QString &password = QString());
+#endif
+
   private:
     QString mFileName;
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+    static QSslKey mSslKey;
+    static QSslCertificate mCertificate;
+    static QList<QSslCertificate> mCaCertificates;
+    static bool mCertificateLoaded;
```
caspermeijn: I am not a fan of static members. Maybe you could use Settings to store this information?

kdwsdl2cpp/common/fileprovider.cpp
```diff
@@ -53,6 +59,23 @@ void FileProvider::cleanUp()
   }
 }
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+bool FileProvider::loadCertificate(const QString &certPath, const QString & password)
+{
+    QFile certFile(certPath);
+    if (certFile.open(QFile::ReadOnly)) {
```
caspermeijn: Would it make sense to report the error messages here? As this function knows what went wrong.

kdwsdl2cpp/src/main.cpp
```diff
@@ -262,6 +286,16 @@ int main(int argc, char **argv)
     Settings::self()->setHelpOnMissing(helpOnMissing);
 
     KWSDL::Compiler compiler;
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+    if (!pkcs12File.isEmpty()) {
+        if (!FileProvider::loadCertificate(pkcs12File, pkcs12Password)) {
+            fprintf(stderr, "Unable to load the %s certificate file\n", pkcs12File.toLocal8Bit().constData());
+            if (!pkcs12Password.isEmpty())
+                fprintf(stderr, "Please make sure that you have passed the correct password");
```
caspermeijn: I think this message is also applicable with a empty password, as an empty password could be the incorrect password.

src/KDSoapClient/moc_predefs.h
```diff
@@ -0,0 +1,389 @@
+#define __SSP_STRONG__ 3
```
caspermeijn: moc files are not supposed to be in version control.

kdwsdl2cpp/src/main.cpp
martonmiklos: Before Qt 5.4 I used QCA for this purpose, but I do not think that we should go down that way.

martonmiklos:
> Thanks for your contribution. Wouldn't it be possible to download the wsdl files from the server before running kdwsdl2cpp? Depending on your use case that may be easier.

A few years ago I did that, however it is painful because the WSDL file has a lot of imports like this one:

> 
<xs:import schemaLocation="https://adatkuldes-teszt.nemzetidohany.gov.hu:8444/ndc/MasterData.svc?xsd=xsd2"

Now I realised that loading certs is so easy, so I think if we could save a couple of minutes for anyone in the future it worth the efforts.

kdwsdl2cpp/common/fileprovider.h
martonmiklos: My motivation behind making these static was to simplify the certificate loading to the fileprovider. If the user pass the certificate I think (due to the nature of the application) it should be expectable to be used in all requests. By using the static members if any developer in the future utilise the Fileprovider the certs will be used automatically, it could not be forgotten. 

What is your objection about the static members? (I am just curious.)

src/KDSoapClient/moc_predefs.h
martonmiklos: S#it. #188  submitted to solve these things forever

kdwsdl2cpp/common/fileprovider.h
caspermeijn: I think that the desired behaviour can be reached by using the Settings singleton. It holds most of the command line settings and I think pkcs12 should be one of them. It probably makes sense to move the load function as well.

I don't have a very concrete reason why, other then that I don't like mixing static and non-static data in the same class definition. My opinion is that a instance of a class should either be self-serving (not dependant on globals, only on references to other classes) or there should only be 1 instance of it (singleton). I am not sure that is a clear explanation. In this case I would prefer one singleton containing all settings and multiple FileProviders holding a reference to the settings (Settings::self() is used in KDSoap).

caspermeijn:
> > Thanks for your contribution. Wouldn't it be possible to download the wsdl files from the server before running kdwsdl2cpp? Depending on your use case that may be easier.
> 
> A few years ago I did that, however it is painful because the WSDL file has a lot of imports like this one:
> `<xs:import schemaLocation="https://*****.gov.hu:8444/ndc/MasterData.svc?xsd=xsd2"`
> Now I realised that loading certs is so easy, so I think if we could save a couple of minutes for anyone in the future it worth the efforts.

Are you aware to these recently added options?

> -import-path <importpath> search for files first in this path before downloading them. may be specified multiple times. the file needs to be located at: <importpath>/<url-host>/<url-path>
> -use-local-files-only     only use local files instead of downloading them automatically. this can be used to force the correctuse of the import-path option

I use that in combination with a [download script](https://gitlab.com/caspermeijn/onvifviewer/blob/master/3rdparty/wsdl/download.sh) to make it possible to build my application offline. Maybe it could be useful to you ;-)

kdwsdl2cpp/common/fileprovider.h
martonmiklos: > I don't have a very concrete reason why, other then that I don't like mixing static and non-static data in the same class definition. 

Thanks for the explanation, totally makes sense, I will fix it soon.

martonmiklos:
> > > Thanks for your contribution. Wouldn't it be possible to download the wsdl files from the server before running kdwsdl2cpp? Depending on your use case that may be easier.
> > 
> > 
> > A few years ago I did that, however it is painful because the WSDL file has a lot of imports like this one:
> > `<xs:import schemaLocation="https://*****.gov.hu:8444/ndc/MasterData.svc?xsd=xsd2"`
> > Now I realised that loading certs is so easy, so I think if we could save a couple of minutes for anyone in the future it worth the efforts.
> 
> Are you aware to these recently added options?
> 
> > -import-path  search for files first in this path before downloading them. may be specified multiple times. the file needs to be located at: //
> > -use-local-files-only     only use local files instead of downloading them automatically. this can be used to force the correctuse of the import-path option
> 
> I use that in combination with a [download script](https://gitlab.com/caspermeijn/onvifviewer/blob/master/3rdparty/wsdl/download.sh) to make it possible to build my application offline. Maybe it could be useful to you ;-)

Thanks for the useful info! (I have missed them.)

kdwsdl2cpp/src/main.cpp
```diff
@@ -67,6 +67,12 @@ static void showHelp(const char *appName)
             "                            use of the import-path option\n"
             "  -help-on-missing          When groups or basic types could not be found, display\n"
             "                            available types (helps with wrong namespaces)\n"
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
+            "  -pkcs12file               Load a certificate from a PKCS12 file. You can use this option\n"
+            "                            if the WSDL file (or files refering to it) is served from a \n"
+            "                            location which require certificate based authentication\n"
+            "  -pkcs12password           Pass the password for the certificate file if required\n"
```
dfaure-kdab: Please add a note that this option is not secure and should be used with caution. Any user on the same machine can use `ps` to see that password as a command-line argument of this process.

I can't think of a better solution though, without adding a dependency on Qt5KeyChain or equivalent

Well, reading the password from a file would already be more secure, since you could limit the permissions for that file to the current user.

martonmiklos: > I can't think of a better solution though, without adding a dependency on Qt5KeyChain or equivalent

We could ask the user for type in the password, if the opening of the cert without password fails.

dfaure-kdab: Having to type the password at every build sounds rather annoying ;)

martonmiklos: Yes indeed. Let's keep it simple by adding a warning to the help.

PR #188, opened 2019-09-05
Extendend gitignore

martonmiklos:
Extended .gitignore to ignore the generated files on a Linux cmake build properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,11 @@
 Makefile
+Makefile.*
+*.pro.user
+*.pro.user*
+*.so
+*.so.*
+*.exe
+moc_*
 *.moc
 *.obj
 *.directory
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -57,3 +57,8 @@ configure.sh
 /unittests/literal_true_false/literal_true_false
 /unittests/msexchange_noservice_wsdl/msexchange_noservice_wsdl
 
+*.o
+*.pro.user
+*.so
+*.so.*
+moc_*
```
caspermeijn: I think these should be at the top with the other build artifacts. It is a bit nit-picky, but I think it makes sense ;-)

martonmiklos: Makes sense, fixed, and added *.exe suffix.

PR #189, opened 2019-10-08
switch foreach to range-based for

anoymouserver:
Since C++11 is already required, the new features could be used as well.


kdwsdl2cpp/src/converter.cpp
```diff
@@ -93,7 +93,7 @@ class TypeCollector
     void registerDerivedClasses()
     {
         XSD::ComplexType::List complexTypes = m_allTypes.complexTypes();
```
dfaure-kdab: Make the list const so that range-for doesn't detach, or use qAsConst if the list is modified later on.

anoymouserver:
I've read the article you mentioned, but qAsConst was added in Qt 5.7, which isn't compatible to
> KD Soap requires Qt 4.7.0 or newer

dfaure-kdab:
Good point. Then we probably want to postpone porting to range-for until we decide to drop Qt4 compatibility.
What prompted you to do this? Did you get compiler warnings due to using foreach? In KDSoap itself or in generated code?

anoymouserver:
No problem. I've only noticed it because I activated `QT_NO_FOREACH` for my local working copy.

I'll make a separate PR for the nullptr commit.

dfaure-kdab:
There's an easy solution to the lack of qAsConst in older Qt versions: rolling our own.
Let's call it kdAsConst.

```
// this adds const to non-const objects (like std::as_const)
template <typename T> Q_DECL_CONSTEXPR typename std::add_const<T>::type &kdAsConst(T &t) noexcept { return t; }                                                                                                                       
// prevent rvalue arguments:
template <typename T> void kdAsConst(const T &&) = delete;
```

winterz:
in the meantime I just applied clang-format coding style and a lot of conflicts now.  Sorry about that.
In master we no longer support Qt4 and required Qt5.7 or higher. Please consider reviving this patch for master with qAsConst

anoymouserver:
I'll rebase it early next week and add the `qAsConst`.

kdwsdl2cpp/src/converter.cpp
```diff
@@ -130,14 +128,14 @@ class TypeCollector
                     usedComplexTypes.insert(typeName, complexType);
 
                     addDependency(complexType.baseTypeName());
-                    Q_FOREACH (const QName &derivedTypeName, complexType.derivedTypes()) {
+                    for (const QName &derivedTypeName : complexType.derivedTypes()) {
```
dfaure-kdab, line 131: FYI:
derivedTypes() returns a QList, so this makes a deep copy. qAsConst() should NOT be used since it's a temporary object. So the "proper" solution is

    const auto derivedTypes = complexType.derivedTypes();

However. This is "just" a code generator, I don't expect the performance difference to be significant, so I'm OK with this going in as is. I however suspect that someone will end up having to make the change anyway, due to warnings from clazy...

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
```diff
@@ -419,7 +419,7 @@ void KDSoapMessageAddressingProperties::readMessageAddressingProperty(const KDSo
         KDSoapMessageRelationship::Relationship relationship;
         relationship.uri = (value.value().toString());
         relationship.relationshipType = addressingNS + QLatin1String("/reply");
-        foreach (KDSoapValue attr, value.childValues().attributes()) {
+        for (KDSoapValue attr : value.childValues().attributes()) {
```
dfaure-kdab, line 422: This one happens at runtime and should therefore be ported to a const local container.

src/KDSoapClient/KDSoapValue.cpp
```diff
@@ -353,7 +353,7 @@ void KDSoapValue::writeChildren(KDSoapNamespacePrefixes &namespacePrefixes, QXml
                                 const QString &messageNamespace, bool forceQualified) const
 {
     const KDSoapValueList &args = childValues();
-    Q_FOREACH (const KDSoapValue &attr, args.attributes()) {
+    for (const KDSoapValue &attr : args.attributes()) {
```
dfaure-kdab, line 356: This one too.

dfaure-kdab (changes requested):

kdwsdl2cpp/src/converter.cpp
anoymouserver: Oh, I didn't knew about this issue with temporary objects in connection with for-range.
This probably affects all occurences of temporary containers here (24 in this commit).

I'll change the lines marked by you .. how about the other 22 ones?

src/KDSoapClient/KDSoapMessageAddressingProperties.cpp
anoymouserver: Actually it does return a `KDSoapValueList &` for `childValues()` and `const QList<KDSoapValue> &` for `attributes()`.
Does the deep copy still occure then?

dfaure-kdab: Ah OK, sorry, I didn't dig into this one. Indeed the ref solves the issue. Returning a const ref is not a great practice in itself, so I didn't think this was happening here.

kdwsdl2cpp/src/converter.cpp
dfaure-kdab: If you can change them all, that's even better :)

PR #192, opened 2019-10-28
add qmake support for soap over udp

apachelogger:
was missing in e0a3bbce1e527f3a5734adb66628b264176faf70 which introduced
soap over udp but neglected to add it to qmake


caspermeijn:
Thanks!

PR #196, opened 2019-12-02
set the cmake install interface directories on the library targets

apachelogger:
the cmake package gets imported targets generated for use by API consumers.

so they can do

```
target_link_libraries(myapp KDSoap::kdsoap)
```

in their cmake code and because KDSoap::kdsoap is an imported target cmake
automatically knows where to find the library. alas, since we didn't
set an interface directory cmake did not know where to find the relevant
headers.
this isn't a problem so long as they are in one of the standard search
paths (e.g. /usr/include), but when the are installed elsewhere (e.g.
/opt/kdsoap/include) they will not be found.

by explicitly setting the INSTALL_INTERFACE to INSTALL_INCLUDE_DIR the
generated KDSoapTargets.cmake will set the relevant path on the target
and thus allow idiomatic use of the imported target without having to
mess with include dirs manually.


Pointedstick:
+1, fixes the issue for me with a self-compiled library.

PR #198, opened 2019-12-13
Add support for WS:Addressing in KDSoapClient

martonmiklos:
If this option is enabled the action parameter (which is passed as an
url parameter by the current implementation) will be put to the message
header. 

I have tested it with a SOAP master implementation written in .NET.

### DIFF
--- a/src/KDSoapClient/KDSoapClientInterface.cpp
+++ b/src/KDSoapClient/KDSoapClientInterface.cpp
@@ -109,7 +109,9 @@ QNetworkRequest KDSoapClientInterfacePrivate::prepareRequest(const QString &meth
         soapHeader += QString::fromLatin1("text/xml;charset=utf-8");
         request.setRawHeader("SoapAction", '\"' + soapAction.toUtf8() + '\"');
     } else if (m_version == KDSoap::SOAP1_2) {
-        soapHeader += QString::fromLatin1("application/soap+xml;charset=utf-8;action=") + soapAction;
+        soapHeader += QString::fromLatin1("application/soap+xml;charset=utf-8");
+        if (m_sendSoapActionInHttpHeader)
+            soapHeader += QString::fromLatin1(";action=") + soapAction;
     }
 
     request.setHeader(QNetworkRequest::ContentTypeHeader, soapHeader.toUtf8());
@@ -135,15 +137,31 @@ QNetworkRequest KDSoapClientInterfacePrivate::prepareRequest(const QString &meth
     return request;
 }
 
-QBuffer *KDSoapClientInterfacePrivate::prepareRequestBuffer(const QString &method, const KDSoapMessage &message, const KDSoapHeaders &headers)
+QBuffer *KDSoapClientInterfacePrivate::prepareRequestBuffer(const QString &method, const KDSoapMessage &message, const QString &soapAction, const KDSoapHeaders &headers)
 {
     KDSoapMessageWriter msgWriter;
     msgWriter.setMessageNamespace(m_messageNamespace);
     msgWriter.setVersion(m_version);
-    const QByteArray data = msgWriter.messageToXml(message, (m_style == KDSoapClientInterface::RPCStyle) ? method : QString(), headers,
-                                                   m_persistentHeaders, m_authentication);
     QBuffer *buffer = new QBuffer;
-    buffer->setData(data);
+    auto setBufferData = [=](const KDSoapMessage &msg) {
+        buffer->setData(msgWriter.messageToXml(msg,
+                                                     (m_style == KDSoapClientInterface::RPCStyle) ? method : QString(),
+                                                     headers, m_persistentHeaders,
+                                                     m_authentication));
+    };
+
+    if (m_sendSoapActionInWsAddressingHeader) {
+        KDSoapMessage messageCopy = message;
+        KDSoapMessageAddressingProperties prop = message.messageAddressingProperties();
+        if (!prop.action().isEmpty())
+            qWarning("Overwriting the action addressing parameter (%s) with the SOAP action (%s)",
+                     prop.action().toLocal8Bit().constData(), soapAction.toLocal8Bit().constData());
+        prop.setAction(soapAction);
+        messageCopy.setMessageAddressingProperties(prop);
+        setBufferData(messageCopy);
+    } else {
+        setBufferData(message);
+    }
     buffer->open(QIODevice::ReadOnly);
     return buffer;
 }
@@ -151,7 +169,7 @@ QBuffer *KDSoapClientInterfacePrivate::prepareRequestBuffer(const QString &metho
 KDSoapPendingCall KDSoapClientInterface::asyncCall(const QString &method, const KDSoapMessage &message, const QString &soapAction,
                                                    const KDSoapHeaders &headers)
 {
-    QBuffer *buffer = d->prepareRequestBuffer(method, message, headers);
+    QBuffer *buffer = d->prepareRequestBuffer(method, message, soapAction, headers);
     QNetworkRequest request = d->prepareRequest(method, soapAction);
     QNetworkReply *reply = d->accessManager()->post(request, buffer);
     d->setupReply(reply);
@@ -181,9 +199,10 @@ KDSoapMessage KDSoapClientInterface::call(const QString &method, const KDSoapMes
     return ret;
 }
 
-void KDSoapClientInterface::callNoReply(const QString &method, const KDSoapMessage &message, const QString &soapAction, const KDSoapHeaders &headers)
+void KDSoapClientInterface::callNoReply(const QString &method, const KDSoapMessage &message,
+                                        const QString &soapAction, const KDSoapHeaders &headers)
 {
-    QBuffer *buffer = d->prepareRequestBuffer(method, message, headers);
+    QBuffer *buffer = d->prepareRequestBuffer(method, message, soapAction, headers);
     QNetworkRequest request = d->prepareRequest(method, soapAction);
     QNetworkReply *reply = d->accessManager()->post(request, buffer);
     d->setupReply(reply);
@@ -332,7 +351,27 @@ void KDSoapClientInterface::setTimeout(int msecs)
     d->m_timeout = msecs;
 }
 
-#ifndef QT_NO_SSL
+bool KDSoapClientInterface::sendSoapActionInHttpHeader() const
+{
+    return d->m_sendSoapActionInHttpHeader;
+}
+
+void KDSoapClientInterface::setSendSoapActionInHttpHeader(bool sendInHttpHeader)
+{
+    d->m_sendSoapActionInHttpHeader = sendInHttpHeader;
+}
+
+bool KDSoapClientInterface::sendSoapActionInWsAddressingHeader() const
+{
+    return d->m_sendSoapActionInWsAddressingHeader;
+}
+
+void KDSoapClientInterface::setSendSoapActionInWsAddressingHeader(bool sendInWsAddressingHeader)
+{
+    d->m_sendSoapActionInWsAddressingHeader = sendInWsAddressingHeader;
+}
+
+#ifndef QT_NO_OPENSSL
 QSslConfiguration KDSoapClientInterface::sslConfiguration() const
 {
     return d->m_sslConfiguration;
--- a/src/KDSoapClient/KDSoapClientInterface.h
+++ b/src/KDSoapClient/KDSoapClientInterface.h
@@ -338,9 +338,49 @@ public:
      */
     void setTimeout(int msecs);
 
+    /**
+     * \brief setSendSoapActionInHttpHeader
+     * \param sendInHttpHeader
+     * This option can be used to enable/disable the sending of the SOAP action parameter
+     * in the Content-Type HTTP header (in the case of SOAP 1.2 requests).
+     * It might be necessary to disable it for certain  SOAP server implementations
+     * in the case if the action is passed in the SOAP envelope's header with WS-Addressing.
+     * \since 1.11
+     */
+    void setSendSoapActionInHttpHeader(bool sendInHttpHeader);
+
+    /**
+     * \brief sendActionInHTTP_Header
+     * \return true if the SOAP action is passed in the HTTP Content-Type of the request.
+     * This parameter is used only in the case of SOAP 1.2 requests.
+     * This option is enabled by default.
+     * \since 1.11
+     */
+    bool sendSoapActionInHttpHeader() const;
+
+    /**
+     * \brief setSendSoapActionInWsAddressingHeader
+     * \param sendInWsAddressingHeader
+     * This function can be used to enable/disable the sending of the SOAP action embedded to
+     * the request's SOAP envelope header. See the headers of the WS-Addressing:
+     * https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/messaging-protocols#message-addressing-headers
+     * \since 1.11
+     */
+    void setSendSoapActionInWsAddressingHeader(bool sendInWsAddressingHeader);
+
+    /**
+     * \brief sendSoapActionInWsAddressingHeader
+     * \return true if the SOAP action is embedded to the request's SOAP envelope
+     * header. See the headers of the WS-Addressing:
+     * https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/messaging-protocols#message-addressing-headers
+     * This option is disabled by default, the SOAP action is sent in the Conten-Type HTTP header.
+     * See setSendHTTPActionParameter for more details.
+     * \since 1.11
+     */
+    bool sendSoapActionInWsAddressingHeader() const;
+
 private:
     friend class KDSoapThreadTask;
-
     KDSoapClientInterfacePrivate *const d;
 };
 
--- a/testtools/httpserver_p.cpp
+++ b/testtools/httpserver_p.cpp
@@ -357,14 +357,41 @@ void HttpServerThread::run()
             break; // normal exit
         }
 
-        // TODO compared with expected SoapAction
         QList<QByteArray> contentTypes = m_headers.value("Content-Type").split(';');
-        if (contentTypes[0] == "text/xml" && m_headers.value("SoapAction").isEmpty()) {
-            qDebug() << "ERROR: no SoapAction set for Soap 1.1";
-            break;
-        } else if (contentTypes[0] == "application/soap+xml" && !contentTypes[2].startsWith("action")) {
-            qDebug() << "ERROR: no SoapAction set for Soap 1.2";
-            break;
+        if (contentTypes[0] == "text/xml") {
+            if (m_headers.value("SoapAction").isEmpty()) {
+                qDebug() << "ERROR: no SoapAction set for Soap 1.1";
+                break;
+            }
+
+            if (!m_expectedSoapAction.isEmpty() && m_headers.value("SoapAction") != m_expectedSoapAction) {
+                qDebug("ERROR: Client sent SoapAction HTTP header (\"%s\") which does not match the expected (\"%s\")",
+                       m_headers.value("SoapAction").constData(), m_expectedSoapAction.constData());
+                break;
+            }
+        } else if (m_clientSendsActionInHttpHeader
+                   && contentTypes[0] == "application/soap+xml") {
+            if (!contentTypes[2].startsWith("action")) {
+                qDebug() << "ERROR: no SoapAction set for Soap 1.2";
+                break;
+            }
+
+            QList<QByteArray> actionParts = contentTypes[2].split('=');
+            if (actionParts.length() < 2) {
+                qDebug("ERROR: The action parameter is malformed in the HTTP Content-type header: \"%s\"",
+                       contentTypes[2].constData());
+                break;
+            }
+
+            if (!m_expectedSoapAction.isEmpty()) {
+                actionParts = actionParts[1].split(';');
+                if (actionParts[0] != m_expectedSoapAction) {
+                    qDebug("ERROR: The 'action' parameter which was sent in the HTTP Content-type header (\"%s\") "
+                           "does not match the expected SOAP action: \"%s\"",
+                           m_headers.value("SoapAction").constData(), m_expectedSoapAction.constData());
+                    break;
+                }
+            }
         }
         lock.unlock();
 
@@ -431,6 +458,26 @@ void HttpServerThread::run()
     }
 }
 
+QByteArray HttpServerThread::expectedSoapAction() const
+{
+    return m_expectedSoapAction;
+}
+
+void HttpServerThread::setExpectedSoapAction(const QByteArray &expectedSoapAction)
+{
+    m_expectedSoapAction = expectedSoapAction;
+}
+
+bool HttpServerThread::clientSendsActionInHttpHeader() const
+{
+    return m_clientSendsActionInHttpHeader;
+}
+
+void HttpServerThread::setClientSendsActionInHttpHeader(bool clientUseWSAddressing)
+{
+    m_clientSendsActionInHttpHeader = clientUseWSAddressing;
+}
+
 const char *KDSoapUnitTestHelpers::xmlEnvBegin11()
 {
     return "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
@@ -441,6 +488,18 @@ const char *KDSoapUnitTestHelpers::xmlEnvBegin11()
            " xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"";
 }
 
+const char *KDSoapUnitTestHelpers::xmlEnvBegin11WithWSAddressing()
+{
+    return  "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+            "<soap:Envelope"
+            " xmlns:soap=\"http://schemas.xmlsoap.org/soap/envelope/\""
+            " xmlns:soap-enc=\"http://schemas.xmlsoap.org/soap/encoding/\""
+            " xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\""
+            " xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\""
+            " xmlns:wsa=\"http://www.w3.org/2005/08/addressing\"";
+}
+
+
 const char *KDSoapUnitTestHelpers::xmlEnvBegin12()
 {
     return "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
@@ -449,6 +508,18 @@ const char *KDSoapUnitTestHelpers::xmlEnvBegin12()
            " xmlns:soap-enc=\"http://www.w3.org/2003/05/soap-encoding\""
            " xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\""
            " xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"";
+}
+
+const char *KDSoapUnitTestHelpers::xmlEnvBegin12WithWSAddressing()
+{
+    return "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+           "<soap:Envelope"
+           " xmlns:soap=\"http://www.w3.org/2003/05/soap-envelope\""
+           " xmlns:soap-enc=\"http://www.w3.org/2003/05/soap-encoding\""
+           " xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\""
+           " xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\""
+           " xmlns:wsa=\"http://www.w3.org/2005/08/addressing\""
+           " xmlns:n1=\"http://www.kdab.com/xml/MyWsdl/\"";
 }
 
 const char *KDSoapUnitTestHelpers::xmlEnvEnd()
--- a/unittests/builtinhttp/builtinhttp.cpp
+++ b/unittests/builtinhttp/builtinhttp.cpp
@@ -407,6 +407,43 @@ private Q_SLOTS:
         }
     }
 
+    // Test if the setClientSendsActionInHttpHeader is set false in the client then the SOAP action
+    // is really not placed into the Content-Type HTTP header
+    void testSetClientSendsActionInHttpHeader()
+    {
+        HttpServerThread server(countryResponse(), HttpServerThread::Public);
+        server.setClientSendsActionInHttpHeader(false);
+
+        qDebug() << "server ready, proceeding" << server.endPoint();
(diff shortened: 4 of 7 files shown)


winterz:
@martonmiklos I'm not the maintainer so I can't comment directly on accepting your patch.

However, before accepting we will need you to fill out a copyright assignment , which you can find in doc/KDSoap-CopyrightAssignmentForm.docx.  You can email the signed copy to me at allen.winter@kdab.com

martonmiklos:
Hi Adam,

Is there any way to fill out the [Please write a brief description of thecontribution] and [Which files have you changed so far, and which new files have you written so far?] sections on page 2 to make the assignment valid for further contributions?
I would like to avoid to having it signed each PR if possible. 

Thanks in advance!

caspermeijn:
Hey @martonmiklos, this sounds useful, however there are some duplicates with other code. 

You can already set WS-Addressing properties using KDSoapMessage::setMessageAddressingProperties. However, you need to do this manually. 
Maybe you could automatically fill the KDSoapMessageAddressingProperties::action when m_useWsAddressing is set.

Why is it needed to remove the "action=" from the SOAP header? Accouring to [RFC3902](https://www.ietf.org/rfc/rfc3902.txt) you could use the action parameter in addition to the WS-Addressing action.

I think useWsAddressing doesn't describe the flag well enough and it is missing documentation in KDSoapClientInterface.h.

martonmiklos:
Hey @caspermeijn 

Many thanks for your feedback!

Actually I was not aware of the setMessageAddressingProperties, so thanks for pointing out, I will modify it. 

However my server implementation requires the removal of the action parameter from the header, at least this is what it suggests in the error reply:
`"Fault 299: Error transferring ... - server replied: Cannot process the message because the content type 'application/soap+xml;charset=utf-8;action=http://localhost/nd/communication/IMasterData/GetProductsSince' was not the expected type 'application/soap+xml; charset=utf-8`

What is your recommendation in this case? I mean in this case my patch should should be modified to allow the user to disable the putting of the action to the header rather than enabling the WS- Addressing.  I know this way we would be a workadound for an erratic server implementation, but in the future we might save some time for another users if we get this implemented.

caspermeijn:
> However the server implementation (what I have to use) requires the removal of the action parameter from the header:
> 
> ```
> "Fault 299: Error transferring ... - server replied: Cannot process the message because the content 
> type 'application/soap+xml;charset=utf-8;action=http://localhost/nd/communication/IMasterData
> /GetProductsSince' was not the expected type 'application/soap+xml; charset=utf-8'
> ```
> 
> What is your recommendation in this case? I mean in this case my patch should should be modified to allow the user to disable the putting of the action to the header rather than enabling the WS- Addressing.

From my point of view the best solution would be to fix the server, but I understand that is not always possible. 

Now I understand why these changes are combined; essentially the action is moved from the HTTP header to the WS-Addressing header. I think this change will be fine with some changes and more documentation.

> 
> I know would be a workaround for an erratic server implementation, but in the future we might save some time for another users if we get this implemented.

martonmiklos:
@caspermeijn I have modified the code to utilise the KDSoapMessage::setMessageAddressingProperties method. I have also added documentation in the KDSoapClientInterface.h.

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -142,21 +144,29 @@ QNetworkRequest KDSoapClientInterfacePrivate::prepareRequest(const QString &meth
     return request;
 }
 
-QBuffer *KDSoapClientInterfacePrivate::prepareRequestBuffer(const QString &method, const KDSoapMessage &message, const KDSoapHeaders &headers)
+QBuffer *KDSoapClientInterfacePrivate::prepareRequestBuffer(const QString &method, KDSoapMessage &message, const QString &soapAction, const KDSoapHeaders &headers)
```
caspermeijn: I think removing const of message is a bad idea. You are now changing the original message, because of the reference.

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -142,21 +144,29 @@ QNetworkRequest KDSoapClientInterfacePrivate::prepareRequest(const QString &meth
...
 {
     KDSoapMessageWriter msgWriter;
     msgWriter.setMessageNamespace(m_messageNamespace);
     msgWriter.setVersion(m_version);
-    const QByteArray data = msgWriter.messageToXml(message, (m_style == KDSoapClientInterface::RPCStyle) ? method : QString(), headers, m_persistentHeaders, m_authentication);
+    if (m_useWsAddressing) {
+        KDSoapMessageAddressingProperties prop;
+        prop.setAction(soapAction);
+        message.setMessageAddressingProperties(prop);
```
caspermeijn: This overwrites any existing AddressingProperties. I think you should get messageAddressingProperties before this.

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -284,6 +295,16 @@ void KDSoapClientInterfacePrivate::setupReply(QNetworkReply *reply)
...
+bool KDSoapClientInterfacePrivate::useWsAddressing() const
+{
+    return m_useWsAddressing;
+}
+
+void KDSoapClientInterfacePrivate::setUseWsAddressing(bool useWsAddressing)
+{
+    m_useWsAddressing = useWsAddressing;
+}
+
```
caspermeijn: I think these duplicate setters are not nessesary, this could be done directly in KDSoapClientInterface::useWsAddressing

src/KDSoapClient/KDSoapClientInterface.h
```diff
@@ -135,7 +135,7 @@ class KDSOAP_EXPORT KDSoapClientInterface
      *  }
      * \endcode
      */
-    KDSoapPendingCall asyncCall(const QString &method, const KDSoapMessage &message,
+    KDSoapPendingCall asyncCall(const QString &method, KDSoapMessage &message,
```
caspermeijn: This interface change is unwanted. It could break existing code,

src/KDSoapClient/KDSoapClientInterface.h
```diff
@@ -346,6 +346,23 @@ class KDSOAP_EXPORT KDSoapClientInterface
...
     void setTimeout(int msecs);
 
+    /**
+     * @brief setUseWsAddressing
+     * @param useWsAddressing
+     * If this parameter is set to true then the outgoing messages
+     * SOAP action parameter will be added as an Action field of the
+     * SOAP envelope instead of passing it as URL parameter.
+     */
+    void setUseWsAddressing(bool useWsAddressing);
```
caspermeijn: I think you should metion WS-Addressing and HTTP headers. And maybe mention the specifications that define it.

src/KDSoapClient/KDSoapClientInterface.cpp
martonmiklos: > Please consider adding a unit test for this new functionality.

Many thanks for the review! I have added an unit test and fixed your mentioned issues.

src/KDSoapClient/KDSoapClientInterface.cpp
martonmiklos: Fixed

src/KDSoapClient/KDSoapClientInterface.cpp
martonmiklos: Good point! Fixed.

src/KDSoapClient/KDSoapClientInterface.h
martonmiklos: Yes i have faced with it when started to work on the unit test. Fixed.

src/KDSoapClient/KDSoapClientInterface.h
martonmiklos: Fixed.

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -142,12 +144,28 @@ QNetworkRequest KDSoapClientInterfacePrivate::prepareRequest(const QString &meth
...
+        data = msgWriter.messageToXml(msg,
+                                                   (m_style == KDSoapClientInterface::RPCStyle) ? method : QString(),
+                                                   headers, m_persistentHeaders,
+                                                   m_authentication);
+    } else {
+        data = msgWriter.messageToXml(message,
+                                                   (m_style == KDSoapClientInterface::RPCStyle) ? method : QString(),
+                                                   headers, m_persistentHeaders,
+                                                   m_authentication);
```
caspermeijn: I think it is not necessary to duplicate this line.

src/KDSoapClient/KDSoapClientInterface.h
```diff
@@ -346,6 +346,26 @@ class KDSOAP_EXPORT KDSoapClientInterface
...
+     */
+    void setUseWS_Addressing(bool useWS_Addressing = true);
+
+    /**
+     * @brief useWS_Addressing
+     * @return
+     * Returns true if the actions of the sent messages are passed in the SOAP envelope's
+     * header section instead of sending them in the request's URL parameters.
+     */
+    bool useWS_Addressing() const;
```
caspermeijn: Personally I don't like the underscore in the name. I would like useWSAddressingAction.

src/KDSoapClient/KDSoapClientInterface.cpp
martonmiklos: Fixed.

src/KDSoapClient/KDSoapClientInterface.h
martonmiklos: Fixed.

src/KDSoapClient/KDSoapClientInterface.h
```diff
@@ -346,6 +346,26 @@ class KDSOAP_EXPORT KDSoapClientInterface
...
+     */
+    void setUseWSAddressing(bool useWSAddressing = true);
+
+    /**
+     * @brief useWSAddressing
+     * @return
+     * Returns true if the actions of the sent messages are passed in the SOAP envelope's
+     * header section instead of sending them in the request's URL parameters.
+     */
+    bool useWSAddressing() const;
```
caspermeijn: Again personally I would make it useWSAddressingAction, as it only affects the action parameter. The other WS-Addressing parameters are not used with this option. But I let others decide whether that is important.

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -142,12 +144,23 @@ QNetworkRequest KDSoapClientInterfacePrivate::prepareRequest(const QString &meth
...
 {
     KDSoapMessageWriter msgWriter;
     msgWriter.setMessageNamespace(m_messageNamespace);
     msgWriter.setVersion(m_version);
-    const QByteArray data = msgWriter.messageToXml(message, (m_style == KDSoapClientInterface::RPCStyle) ? method : QString(), headers, m_persistentHeaders, m_authentication);
+
+    KDSoapMessage msg = message;
+    if (m_useWSAddressing) {
+        KDSoapMessageAddressingProperties prop = message.messageAddressingProperties();
+        prop.setAction(soapAction);
```
dfaure-kdab: Until now it was up to the caller to configure the addressing properties correctly. Now it's done magically in here, but do we really want that? This would overwrite any action set by the caller.

In other words this patch combines two features:
 - if you don't want the action http parameter, call setUseWSAddressing(true)
 - oh but when you do that, we'll overwrite your earlier call to messageAddressingProperties().setAction(), if any.

It seems more flexible to me to separate both things, especially if, as this patch shows, we end up needing more flexibility than reading the spec initially indicated.

I admit that I don't know how all this interacts with kdwsdl2cpp though, but whatever we decide, we can always adapt the code generator accordingly, if needed.

src/KDSoapClient/KDSoapClientInterface.h
dfaure-kdab: Ah, the usual issue. Naming a method after what it does, or after when it should be used. The latter might make sense when one wants to hide some details about what it does, but in general that's a slippery slope... like here.

After reading https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/messaging-protocols I believe this should indeed be renamed, because it *is* possible to use the action HTTP parameter even when using WS-Addressing, as long as it matches the action specified in the XML -- it's an optimization so that message dispatch doesn't need to parse the XML.

Therefore the logic here should not be "if you're using WSAddressing, tell us and we'll skip the action parameter".
It should simply be "Do you want the action parameter emitted? Say no if your server is broken or if you're too lazy to set it to the right value."

The method name shouldn't even refer to WSAddressing, the action parameter exists independently from that (on the contrary, it mostly makes sense when WSAddressing is NOT used).

Therefore I suggest something like `setSendHTTPActionParameter(bool)`. With an explanation in the documentation that matches the one on the above webpage [unless you find contradictory information elsewhere....].

(The trick is to not confuse SOAP-1.1 users, this does not affect the SOAPAction HTTP header).
setSendSoap12HTTPContentTypeActionParameter becomes a bit long :-)

src/KDSoapClient/KDSoapClientInterface.cpp
caspermeijn: Yeah, it probably makes sense to split the two up. And looking at some of the code generated by kdwsdl2cpp, it could make sense to let it generate the code for writing the action in the KDSoapMessageAddressingProperties. It already generates the KDSoapMessage and has the action string available.

martonmiklos: My wsdl file contains a  <wsaw:UsingAddressing/> which (I think) refers to the usage of the WS-Addressing:

```
<?xml version="1.0" encoding="utf-8"?>
<wsdl:definitions xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/" xmlns:wsx="http://schemas.xmlsoap.org/ws/2004/09/mex" xmlns:i0="http://localhost/nd/communication/" xmlns:wsu="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd" xmlns:wsa10="http://www.w3.org/2005/08/addressing" xmlns:wsp="http://schemas.xmlsoap.org/ws/2004/09/policy" xmlns:wsap="http://schemas.xmlsoap.org/ws/2004/08/addressing/policy" xmlns:msc="http://schemas.microsoft.com/ws/2005/12/wsdl/contract" xmlns:soap12="http://schemas.xmlsoap.org/wsdl/soap12/" xmlns:wsa="http://schemas.xmlsoap.org/ws/2004/08/addressing" xmlns:wsam="http://www.w3.org/2007/05/addressing/metadata" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:tns="http://tempuri.org/" xmlns:soap="http://schemas.xmlsoap.org/wsdl/soap/" xmlns:wsaw="http://www.w3.org/2006/05/addressing/wsdl" xmlns:soapenc="http://schemas.xmlsoap.org/soap/encoding/" name="MasterData" targetNamespace="http://tempuri.org/">
  <wsp:Policy wsu:Id="WSHttpBinding_IMasterData_policy">
    <wsp:ExactlyOne>
      <wsp:All>
        <sp:TransportBinding xmlns:sp="http://schemas.xmlsoap.org/ws/2005/07/securitypolicy">
          <wsp:Policy>
            <sp:TransportToken>
              <wsp:Policy>
                <sp:HttpsToken RequireClientCertificate="true"/>
              </wsp:Policy>
            </sp:TransportToken>
            <sp:AlgorithmSuite>
              <wsp:Policy>
                <sp:Basic256/>
              </wsp:Policy>
            </sp:AlgorithmSuite>
            <sp:Layout>
              <wsp:Policy>
                <sp:Strict/>
              </wsp:Policy>
            </sp:Layout>
          </wsp:Policy>
        </sp:TransportBinding>
        <wsaw:UsingAddressing/>
```
I agree with both of you about separating the two things.

src/KDSoapClient/KDSoapClientInterface.h
martonmiklos: In the WSDL what the server provides I have a
`<wsaw:UsingAddressing/>` tag while the linked docs mentions the `<wsap:UsingAddressing />` for turning on the WS-Addressing 2004/08.

In my case the kdwsdl2cpp generated code sent the correct action in the HTTP parameter what is in the WSDL's soap12:address field location attribute, so the R2221 condition of the mentioned spec was satisfied before this PR too:

Error message while action is in the header:

```
"Fault 299: Error transferring ... - server replied: Cannot process the message because the content 
type 'application/soap+xml;charset=utf-8;action=http://localhost/nd/communication/IMasterData
/GetProductsSince' was not the expected type 'application/soap+xml; charset=utf-8'
```

Relevant WSDL part:

```
<wsdl:operation name="GetProductsSince">
      <soap12:operation soapAction="http://localhost/nd/communication/IMasterData/GetProductsSince" style="document"/>
      <wsdl:input>
        <soap12:body use="literal"/>
      </wsdl:input>
      <wsdl:output>
        <soap12:body use="literal"/>
      </wsdl:output>
      <wsdl:fault name="NDFaultContractFault">
        <soap12:fault name="NDFaultContractFault" use="literal"/>
      </wsdl:fault>
    </wsdl:operation>
```

And I got the same fault even if the request's wsa:Action is correct, so my server implementation seems to be ignoring R2222 as well:

```
<?xml version="1.0" encoding="UTF-8"?><soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope" xmlns:soap-enc="http://www.w3.org/2003/05/soap-encoding" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:wsa="http://www.w3.org/2005/08/addressing" xmlns:n1="http://localhost/nd/communication/"><soap:Header><wsa:Action>http://localhost/nd/communication/IMasterData/GetProductsSince</wsa:Action></soap:Header><soap:Body><n1:GetProductsSince><n1:shopId>a529a279-dc92-4943-93b8-d2912941381b</n1:shopId><n1:modifiedSince>2019-11-21T10:59:22.218</n1:modifiedSince></n1:GetProductsSince></soap:Body></soap:Envelope>
```

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -142,21 +144,37 @@ QNetworkRequest KDSoapClientInterfacePrivate::prepareRequest(const QString &meth
...
     msgWriter.setVersion(m_version);
-    const QByteArray data = msgWriter.messageToXml(message, (m_style == KDSoapClientInterface::RPCStyle) ? method : QString(), headers, m_persistentHeaders, m_authentication);
+
     QBuffer *buffer = new QBuffer;
-    buffer->setData(data);
+    if (m_sendSOAPActionInWSAAction) {
+        KDSoapMessage messageCopy = message;
+        KDSoapMessageAddressingProperties prop = message.messageAddressingProperties();
+        if (!prop.action().isEmpty())
+            qWarning() << "Overwriting the action addressing parameter with the SOAP action";
```
dfaure-kdab: I suggest printing both values, so the developer doing this gets an idea of where the problem comes from.

src/KDSoapClient/KDSoapClientInterface.h
```diff
@@ -346,9 +346,46 @@ class KDSOAP_EXPORT KDSoapClientInterface
...
     void setTimeout(int msecs);
 
+    /**
+     * @brief setSendHTTPActionParameter
+     * @param sendHTTPAction
+     * This option can be used to enable/disable the sending of the SOAP action parameter
+     * in the Content-Type HTTP header (in the case of SOAP 1.2 requests).
+     * It might be necessary to disable it for certain  SOAP server implementations
+     * in the case if the action is passed in the SOAP envelope's header with WS-Addressing.
+     */
```
dfaure-kdab: Missing `@since 1.9` for all 4 new methods.

src/KDSoapClient/KDSoapClientInterface.h
```diff
@@ -346,9 +346,46 @@ class KDSOAP_EXPORT KDSoapClientInterface
...
+    void setSendHTTPActionParameter(bool sendHTTPAction = true);
```
dfaure-kdab: I'd remove the default value, since the member defaults to true, so calling `setSendHttpActionParameter()` basically does nothing.

src/KDSoapClient/KDSoapClientInterface.h
```diff
@@ -346,9 +346,46 @@ class KDSOAP_EXPORT KDSoapClientInterface
...
+
+    /**
+     * @brief sendHTTPActionParameter
+     * @return
```
dfaure-kdab: join with next line (`@return true if ....`)

testtools/httpserver_p.cpp
```diff
@@ -397,13 +397,15 @@ void HttpServerThread::run()
         }
 
         // TODO compared with expected SoapAction
```
dfaure-kdab: bonus points for implementing the TODO while at it ;)

unittests/builtinhttp/builtinhttp.cpp
```diff
@@ -426,6 +426,38 @@ private Q_SLOTS:
         }
     }
 
+    // Test that that if the setUseWSAddressing is set int the client  then the SOAP action
```
dfaure-kdab: `that that`? ;)

`int` -> `in`

two spaces before "then"

unittests/builtinhttp/builtinhttp.cpp
```diff
@@ -426,6 +426,38 @@ private Q_SLOTS:
...
+    // is really placed to the SOAP envelope's header and not embedded to the  Content-Type HTTP header
```
dfaure-kdab: I'd say `to` -> `in` both times

two spaces before Content-Type

unittests/builtinhttp/builtinhttp.cpp
```diff
@@ -426,6 +426,38 @@ private Q_SLOTS:
...
+
+        qDebug() << "server ready, proceeding" << server.endPoint();
+        KDSoapClientInterface client(server.endPoint(), countryMessageNamespace());
+        client.setSendHTTPActionParameter();
+
+        client.setSoapVersion(KDSoapClientInterface::SOAP1_1);
+        QString action = QStringLiteral("http://localhost/getEmployeeCountry");
+        KDSoapPendingCall call = client.asyncCall(QStringLiteral("getEmployeeCountry"), countryMessage(), action);
+        QVERIFY(!call.isFinished());
+        QTest::qWait(1000);
```
dfaure-kdab: This makes the test really slow. Better wait "just the right amount".
One easy solution is to just use call() instead of asyncCall().

unittests/builtinhttp/builtinhttp.cpp
```diff
@@ -426,6 +426,38 @@ private Q_SLOTS:
...
+        QVERIFY(xmlBufferCompare(server.receivedData(), expectedCountry11WithWSAddressingRequest()));
+        QCOMPARE(server.header("Content-Type").constData(), "text/xml;charset=utf-8");
+
+        client.setSoapVersion(KDSoapClientInterface::SOAP1_2);
+        call = client.asyncCall(QStringLiteral("getEmployeeCountry"), countryMessage(), action);
```
dfaure-kdab: (same)

unittests/builtinhttp/builtinhttp.cpp
```diff
@@ -426,6 +426,38 @@ private Q_SLOTS:
...
+    // Test that that if the setUseWSAddressing is set int the client  then the SOAP action
+    // is really placed to the SOAP envelope's header and not embedded to the  Content-Type HTTP header
+    void testUseWSAddressingCall()
+    {
+        HttpServerThread server(countryResponse(), HttpServerThread::Public);
+        server.setClientUseWSAddressing();
+
+        qDebug() << "server ready, proceeding" << server.endPoint();
+        KDSoapClientInterface client(server.endPoint(), countryMessageNamespace());
+        client.setSendHTTPActionParameter();
```
dfaure-kdab: That's the default, so this line is a no-op.

What is actually missing is a test for setSendHTTPActionPameter(false), isn't it?

src/KDSoapClient/KDSoapClientInterface.cpp
martonmiklos: Indeed, and I will also supress the warning if no overwrite is done (when the SOAP action and the action is identical).

src/KDSoapClient/KDSoapClientInterface.h
martonmiklos: I have changed the placeholders from @param format to \param to be consistent with the previous method comments.

src/KDSoapClient/KDSoapClientInterface.h
martonmiklos: Makes sense.

testtools/httpserver_p.cpp
martonmiklos: Done

unittests/builtinhttp/builtinhttp.cpp
martonmiklos: Fixed

unittests/builtinhttp/builtinhttp.cpp
martonmiklos: Fixed

unittests/builtinhttp/builtinhttp.cpp
martonmiklos: I agree, fixed.

unittests/builtinhttp/builtinhttp.cpp
martonmiklos: Fixed

unittests/builtinhttp/builtinhttp.cpp
martonmiklos: Yes, actually I submitted to review before fixing the unit tests. Now I think it should be ok including the tests.

winterz:
this one has been sitting for a very long time.
any interest in reviving it for the next kdsoap release which I'd like to get out the door very soon

martonmiklos:
> this one has been sitting for a very long time.
> any interest in reviving it for the next kdsoap release which I'd like to get out the door very soon

Thanks for the heads up, I will look into this.

src/KDSoapClient/KDSoapClientInterface.h
martonmiklos: Fixed

martonmiklos:
Hello folks!

I dusted off this a bit. 

Changes I made today:
 * Fixed unittests: there were some QByteArray - QString comparisons which failed to compile with 5.15. 
 * Rewritten the commit message to reflect what we ended up actually implementing :)
 * Renamed the added functions to have an underscore between the uppercase acronyms and the next uppercase word: setSendSOAP_ActionInHTTP_Header and setSendSOAP_ActionInWS_AddressingHeader and made them consistent naming wise.
 
I built locally with unit tests and get the relevant tests passing.
If Travis and appveyor does not find any issues while I sleep, I would like to ask @dfaure-kdab to do a review.

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -135,23 +137,40 @@ QNetworkRequest KDSoapClientInterfacePrivate::prepareRequest(const QString &meth
...
-    buffer->setData(data);
+    if (m_sendSOAP_ActionInWS_AddressingHeader) {
+        KDSoapMessage messageCopy = message;
+        KDSoapMessageAddressingProperties prop = message.messageAddressingProperties();
+        if (!prop.action().isEmpty())
+            qWarning("Overwriting the action addressing parameter (%s) with the SOAP action (%s)",
+                     prop.action().toLocal8Bit().constData(), soapAction.toLocal8Bit().constData());
+        prop.setAction(soapAction);
+        messageCopy.setMessageAddressingProperties(prop);
+        buffer->setData(msgWriter.messageToXml(messageCopy,
```
dfaure-kdab, line 156: This 4-lines statement could be a local lambda, to avoid the code duplication. It would ease future maintenance.

unittests/builtinhttp/builtinhttp.cpp
```diff
@@ -407,6 +407,43 @@ private Q_SLOTS:
...
+        client.setSoapVersion(KDSoapClientInterface::SOAP1_2);
+        QString action = QStringLiteral("http://localhost/getEmployeeCountry");
+        server.setExpectedSoapAction(action.toLocal8Bit());
+        KDSoapMessage ret= client.call(QStringLiteral("getEmployeeCountry"), countryMessage(), action);
+        QVERIFY(!ret.isFault());
+        QVERIFY(xmlBufferCompare(server.receivedData(), expectedCountryRequest12()));
+        QCOMPARE(server.header("Content-Type").constData(), "application/soap+xml;charset=utf-8");
+    }
+
+    // Test if the setSendSOAPActionInWSAAction is set true in the client then the SOAP action
```
dfaure-kdab, line 429: The method name doesn't match anymore.

src/KDSoapClient/KDSoapClientInterface.cpp
```diff
@@ -332,7 +352,27 @@ void KDSoapClientInterface::setTimeout(int msecs)
     d->m_timeout = msecs;
 }
 
-#ifndef QT_NO_SSL
+bool KDSoapClientInterface::sendSOAP_ActionInHTTP_Header() const
```
dfaure-kdab, line 355: I don't want to nitpick too much, but I have to say the underscores after acronyms surprise me very much. My brain splits at underscores first, and then reads the 3 substrings, it doesn't really work.

Qt uses another solution: treating acronyms like normal words, i.e. with titlecase (for example the Http in "QNetworkProxy::HttpCachingProxy"). This would give

    sendSoapActionInHttpHeader()

Isn't this much more readable?
And consistent with Qt, which is an important point for a Qt-based library.

dfaure-kdab (changes requested):
Thanks!

src/KDSoapClient/KDSoapClientInterface.cpp
martonmiklos: Done

unittests/builtinhttp/builtinhttp.cpp
martonmiklos: Fixed

src/KDSoapClient/KDSoapClientInterface.cpp
martonmiklos: > I don't want to nitpick too much
Hah, nitpicking is never too much.

>  And consistent with Qt, which is an important point for a Qt-based library.
I was not aware of this Qt coding style, but I will keep myself to this practice in the future.

unittests/builtinhttp/builtinhttp.cpp
martonmiklos: Fixed

PR #200, opened 2020-01-23
[kdwsdl2cpp]Do not overwrite output files in the case their content unchanged

martonmiklos:
Mainly useful when tweaking kdwsdl2cpp because the generated files not
gooing to be modified if their contents did not changed. This could
reduce the compilation time of the projects including the generated
codes.

This change have been manually copied over from the kode's libkode where
it got reviewed by @dfaure-kdab:
https://github.com/cornelius/kode/pull/32


dfaure-kdab:
Thanks!

PR #204, opened 2020-02-24
Utilize libkode from https://github.com/cornelius/libkode

### LINKED ISSUE #201: libkode uniformization with kode
I have opened this issue to discuss if it would be useful to bring the two copies of libkode (one in KDSoap and one in the kode) under one hood.  I have some ideas to implement in kode which might be beneficial to add to the KDSoap  as well.

My recommendation would be the following:
- Agree on a common coding style
- Reformat the KDSoap's libkode according to that format
- Move the the libkode from KDSoap to a separate git repository under the governance of KDAB (since it is business critical here), and add it to the KDSoap as a git submodule
- I will replace the kode's libkode with that repo and work on that until it work with it as a submodule.  It will result some PRs to the common repo, which will needs some reviewing resources. 

I know it is going to be a tedious task, and will require efforts on KDAB side as well but manually merging kode changes would be tedious as well ;).

Thank you for your feedback in advance!

martonmiklos:
Fixes #201

I did an md5sum comparison on the kdwsdl2cpp built from the current master and the fork:
https://gist.github.com/martonmiklos/727604c5a558a1d2cee4785f3fde79ec
and they are match (so the clang-formatter did good job).

I have not seen any public CI setup in the project (if I am not mistaken), but I assume you have some internal. 

Please keep in mind that some build scripts tweaks might be necessary: the submodule needs to be initialized and updated after checkout.


dfaure-kdab:
KDSoap/unittests/list_restriction/test_list_restriction.cpp:26:10: fatal error: messagehandler.h: No such file or directory
(with cmake). Investigating

dfaure-kdab:
OK, fixed in commit e6d4441.

martonmiklos:
I have missed running tests after the submodule conversion, I apologize for that.

caspermeijn:
It is great to see the merging of these two code bases.

I think there needs to be a changelog enty for this.

Is it possible for CMake to retrieve the submodule automatically, otherwise I would like an instruction in the README. (One of my CI jobs fail because of missing files on the master branch ;-)

martonmiklos:
I agree with the changelog entry. 

> Is it possible for CMake to retrieve the submodule automatically, otherwise I would like an instruction in the README. (One of my CI jobs fail because of missing files on the master branch ;-)

Well it would bring up some questions because while on Linux git used to be in the path on Windows it could be found in several folders. From the limited examples where I saw submodules usually the CI jobs do a git submodule init and update after the clone.

dfaure-kdab:
Yeah, to each tool its job. I agree that this should just be done by the CI job, just like the git clone itself.
In one of our internal projects that uses cmake+git+submodules, we simply added a check like: if the submodule path doesn't exist, `message(FATAL_ERROR "Please do git submodule update --init --recursive")`. This helps humans, not so much CI jobs :-)

caspermeijn:
My CI job is for an application that uses KDSoap. Until the 1.9 release it needed master. My recent build failed because of this merge to master. I knew of this change happening, therefore I understood the problem. But if you dont know, it can become a long the debugging session.

I think the CMake FATAL would help a lot.

martonmiklos:
@caspermeijn @dfaure-kdab  many thanks for the ideas, I have just submitted a PR #205  which makes both build systems fail in the case of missing submodule.

PR #211, opened 2020-03-13
fixed missing set

### LINKED ISSUE #208: 1.9.0 release - LICENSE.US.txt file missing
Trying to run configure on Windows and get the following error:

'license' is not recognized as an internal or external command,
operable program or batch file.

In my case it is hitting the else branch of line 33 of configure.bat:
if exist %SOURCE_DIR%\LICENSE.US.txt (

The LICENSE.US.txt file is missing.

anoymouserver:
fixes #208

I would recommend to revise the configure script because of the removed US license. Currently it is only possible to select LGPL.
Also the `autogen.py` auto-accepts a license without showing anything but running it is required to generate the `configure.bat` / `configure.sh`

### DIFF
--- a/autogen/configure.bat.in
+++ b/autogen/configure.bat.in
@@ -40,17 +40,17 @@ if exist %SOURCE_DIR%\LICENSE.@PRODUCT_LICENSE_FREE@.txt (
             echo Type 3 for the %Product_Space% Commercial License for anywhere outside USA/Canada.
             echo Anything else cancels.
             echo.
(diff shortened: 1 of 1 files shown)


winterz:
I combined your fixes along with other fixes that handles the removal of LICENSE.US.txt
see e42142f56a2f4d7df7f2f35558df4d5815bb8f06 in kdsoap

thanks for the help.  no longer any need for this pull request.

PR #214, opened 2020-03-29
Remove dependency on Settings from the libkode's FileProvider

martonmiklos:
- [libkode]Remove dependency on Settings from the libkode's FileProvider
- [kdwsdl2cpp]Make the SSL configuration global if PKCS12 certificate
loaded

Before merging the following PR in libkode needs to be merged:
https://github.com/cornelius/libkode/pull/10


kdwsdl2cpp/src/main.cpp
```diff
@@ -290,8 +293,38 @@ int main(int argc, char **argv)
...
 #if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
     if (!pkcs12File.isEmpty()) {
-        if (!Settings::self()->loadCertificate(pkcs12File, pkcs12Password))
-            return -1;
+        QFile certFile(pkcs12File);
+        if (certFile.open(QFile::ReadOnly)) {
+            QSslKey key;
+            QSslCertificate certificate;
+            QList<QSslCertificate> caCertificates;
+            auto mCertificateLoaded = QSslCertificate::importPkcs12(&certFile,
```
dfaure-kdab: The naming needs to be adapted (no 'm' prefix) --> const bool certificateLoaded =

martonmiklos: Fixed, thanks for spotting it!

dfaure-kdab:
You also need to update the submodule in order to get the libkode commit, right?

martonmiklos:
> You also need to update the submodule in order to get the libkode commit, right?

Well I think it is not necessary. The libkode in this PR points to e588f5d007a5c5bae25c2010561bb90f4f2f59ae which is now available in the upstream libkode (thanks to you merge):
https://github.com/cornelius/libkode/commit/e588f5d007a5c5bae25c2010561bb90f4f2f59ae
Or do you prefer to point to the merge commit?

dfaure-kdab:
Oh I hadn't seen the "Submodule libkode updated" in this commit. That's exactly what I meant, which is fine.

PR #225, opened 2020-07-27
Fix #224

### LINKED ISSUE #224: [kdwsdl2cpp]Allow generating only sync/async client implementation
I was wondering if it would be useful to add a command line switch to the kdwsdl2cpp which would allow the users to generate client codes which contains only synchronous or asynchronous implementations. 

This could result smaller generated code, and would prevent calling a sync call by a mistake instead of an async (like I did just now). 

If you like the idea please let me know, it looks to be a low hanging fruit to implement for me.

martonmiklos:
Add command line options to turn off generation of the sync/async/asyncjob API to the client code.

### DIFF
--- a/kdwsdl2cpp/src/main.cpp
+++ b/kdwsdl2cpp/src/main.cpp
@@ -76,8 +76,11 @@ static void showHelp(const char *appName)
             "                            location which require certificate based authentication\n"
             "  -pkcs12password           Pass the password for the certificate file if required.\n"
             "                            This option is not secure and should be used with caution\n"
-            "                            if other users of the machine are capable to see the running "
+            "                            if other users of the machine are capable to see the running\n"
             "                            processes ran by the current user.\n"
+            "  -no-sync                  Do not generate synchronous API methods to the client code\n"
+            "  -no-async                 Do not generate asynchronous API methods to the client code\n"
+            "  -no-async-jobs            Do not generate asynchronous job API classes to the client code\n"
 #endif
             "\n", appName, appName, appName);
 }
@@ -104,6 +107,7 @@ int main(int argc, char **argv)
     QStringList importPathList;
     bool useLocalFilesOnly = false;
     bool helpOnMissing = false;
+    bool skipAsync = false, skipSync = false, skipAsyncJobs = false;
 #if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0) && !defined(QT_NO_SSL)
     QString pkcs12File, pkcs12Password;
 #endif
@@ -240,6 +244,12 @@ int main(int argc, char **argv)
             }
             pkcs12Password = QLatin1String(argv[arg]);
 #endif
+        } else if (opt == QLatin1String("-no-sync")) {
+            skipSync = true;
+        } else if (opt == QLatin1String("-no-async")) {
+            skipAsync = true;
+        } else if (opt == QLatin1String("-no-async-jobs")) {
+            skipAsyncJobs = true;
         } else if (!fileName) {
             fileName = argv[arg];
         } else {
@@ -293,6 +303,9 @@ int main(int argc, char **argv)
     Settings::self()->setImportPathList(importPathList);
     Settings::self()->setUseLocalFilesOnly(useLocalFilesOnly);
     Settings::self()->setHelpOnMissing(helpOnMissing);
+    Settings::self()->setSkipSync(skipSync);
+    Settings::self()->setSkipAsync(skipAsync);
+    Settings::self()->setSkipAsyncJobs(skipAsyncJobs);
 
     KWSDL::Compiler compiler;
 #if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0) && !defined(QT_NO_SSL)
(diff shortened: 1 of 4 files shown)


kdwsdl2cpp/src/main.cpp
```diff
@@ -76,8 +76,11 @@ static void showHelp(const char *appName)
             "                            location which require certificate based authentication\n"
             "  -pkcs12password           Pass the password for the certificate file if required.\n"
             "                            This option is not secure and should be used with caution\n"
-            "                            if other users of the machine are capable to see the running "
+            "                            if other users of the machine are capable to see the running\n"
             "                            processes ran by the current user.\n"
+            "  -no-sync                  Do not generate syncronous API methods to the client code\n"
```
dfaure-kdab, line 81: Typo: synchronous with a 'h' (same on the next two lines)

kdwsdl2cpp/src/main.cpp
```diff
@@ -76,8 +76,11 @@ static void showHelp(const char *appName)
...
+            "  -no-async                 Do not generate asyncronous API methods to the client code\n"
+            "  -no-async-jobs            Do not generate asyncronous job API methods to the client code\n"
```
dfaure-kdab, line 83: Those are classes, not just methods. I suggest s/methods/classes/

dfaure-kdab:
It would also be very interesting to use these options in all the unittest, depending on what each of them needs, to reduce compilation times; but this can be done separately or left as an exercise to the reader ;)

martonmiklos:
> It would also be very interesting to use these options in all the unittest, depending on what each of them needs, to reduce compilation times; but this can be done separately or left as an exercise to the reader ;)

Which one do you prefer? I mean between fixing this PR or in a separate.

For the qmake unittests I will need add KDWSDL_OPTIONS += -no-async/sync/asyncjob right?
For cmake set(KSWSDL2CPP_OPTION ${KSWSDL2CPP_OPTION} -no-async/sync/asyncjob) should do the trick.

dfaure-kdab:
> Which one do you prefer? I mean between fixing this PR or in a separate PR option.

Doesn't really matter, but let's say separate, this way we can get this one in earlier.

> For the qmake unittests I will need add KDWSDL_OPTIONS += -no-async/sync/asyncjobs right?

Right, just like e.g. unittests/optionaltype_boost_optional/optionaltype_boost_optional.pro does.

> For cmake set(KSWSDL2CPP_OPTION ${KSWSDL2CPP_OPTION} -no-async/sync/asyncjobs) should do the trick.

Yes (space-separated if more than one).

Not just unittests/ but also the few examples/.

Thanks!

dfaure-kdab:
Added to changelog in commit 42faf46

martonmiklos:
> > Which one do you prefer? I mean between fixing this PR or in a separate PR option.
> 
> Doesn't really matter, but let's say separate, this way we can get this one in earlier.
> 
> > For the qmake unittests I will need add KDWSDL_OPTIONS += -no-async/sync/asyncjobs right?
> 
> Right, just like e.g. unittests/optionaltype_boost_optional/optionaltype_boost_optional.pro does.
> 
> > For cmake set(KSWSDL2CPP_OPTION ${KSWSDL2CPP_OPTION} -no-async/sync/asyncjobs) should do the trick.
> 
> Yes (space-separated if more than one).
> 
> Not just unittests/ but also the few examples/.
> 
> Thanks!



> > Which one do you prefer? I mean between fixing this PR or in a separate PR option.
> 
> Doesn't really matter, but let's say separate, this way we can get this one in earlier.
> 
> > For the qmake unittests I will need add KDWSDL_OPTIONS += -no-async/sync/asyncjobs right?
> 
> Right, just like e.g. unittests/optionaltype_boost_optional/optionaltype_boost_optional.pro does.
> 
> > For cmake set(KSWSDL2CPP_OPTION ${KSWSDL2CPP_OPTION} -no-async/sync/asyncjobs) should do the trick.
> 
> Yes (space-separated if more than one).
> 
> Not just unittests/ but also the few examples/.
> 
> Thanks!

Alright, it was a really good point that you highlighted the updating of tests (and in general test/validate the change itself): I started to work on them and it turned out that it is possible to generate non compliable code: the date_example missed the QDate include.

dfaure-kdab:
Good catch.

Hmm, BTW, we're thinking more and more about dropping the qmake-based buildsystem. What are you thoughts on that?
Maybe it's not worth you putting time into it, cmake would be good enough.

martonmiklos:
> Hmm, BTW, we're thinking more and more about dropping the qmake-based buildsystem. What are you thoughts on that?
> Maybe it's not worth you putting time into it, cmake would be good enough.

@dfaure 
To respond to this question: 

I have a single project where I use KDSoap at the moment and I build it with qmake. I am using my fork in that project until we manage to upstream #198 . Most of my contributions (apart from the Kode related ones) to KDSoap was derived from working on that project and now it's client side is rolling solidly now, and I do not expect big changes on the server side in the near future. I can live with kind of clear conscience if it will rely on my fork (ie. if we merge #198  after the qmake support got dropped).

I understand and feel the pain which resulted by the maintaining two parallel build systems and sooner or later I will need to familiarise myself to cmake. 

I have updated the unittest and examples qmake project files too because it was easy this time (it was way more time to figure out which API generation could be dropped from each unittest than adding the proper line to the qmake projects). And I did not wanted to be the first one who introduce the first cmake only feature here ;).

dfaure-kdab:
I'd rather not do something that pushes you towards using a fork ;)
Let me know when you've switched over to cmake, it's so much better than qmake (which BTW is moving to the sidelines in Qt 6)  :-)

PR #235, opened 2021-02-08
Add unit test showing failure at generating code for enum with length…

raphaelcotty:
The patch adds a new unit test (enum_with_length_restriction) which shows how the code generator fails with an enumeration type with a length restriction.


dfaure-kdab:
Cherry-picked as commit e9fbfa6f63c9474ab1543fc6f0a70b484cdb0529

PR #237, opened 2021-02-10
Update unit test showing failure at generating code for annotation wi…

raphaelcotty:
…th length restriction


dfaure-kdab:
Thanks for the unittest extension. The commit log is misleading, annotations are unrelated, this is about the fact that it's a restriction with patterns and not like an enum. Commit log amended, commit pushed, as well as the fix: commit 3a7e89b.

PR #243, opened 2021-05-27
ignore CMake build folder

anoymouserver:
The directory is suggested here but not ignored by git..
https://github.com/KDAB/KDSoap/blob/dc4754f03aebb3201e61742c515df808f77471e6/INSTALL.txt#L19-L25

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,8 @@ configure.bat
 configure.sh
 .qmake.stash
 
+/build
+
 /include
 
 /examples/holidays_gui/qrc_resources.cpp


winterz:
I don't think so.  how about your global $HOME/.gitignore?

anoymouserver:
Yes it could. But if someone clones this repo and builds it as described in the INSTALL.txt, it should ignore the build results per default in my opinion. 🤔

dfaure-kdab:
This makes sense to me, it's indeed consistent with the installation instructions. It's not like we have to ignore all possible names people might come up with (I have build_qt5_cmake, build_qt5_qmake, build_qt6 etc.) ;)
Allen: was that the basis for your objection?

winterz:
no major objection.   I use "build" dirs in so many projects that it makes more sense to me to use global .gitignore

PR #255, opened 2023-02-28
Some server implementations like the one in use by …

tizbac:
www.ilportaledellautomobilista.it are broken and fail if Password is passed before Username
Also Sending Nonce and Created to those implementations will cause authentication to fail


tizbac:
Wondering also if Nonce and Created could be simply be removed when not using password digest mode

Issue #229, opened 2020-08-03
Qmake error including kdsoap.pri 


tomas4TW:
Hello,
I copied and included kdsoap.pri into my project, as was mentioned in documentation.
`#KD-SOAP`
`KDWSDL_OPTIONS = -server          #server`
`KDSOAPDIR = $$KDSOAP_PATH    #or enviroment variable`
`WSDL_DIR = generated`
`KDWSDL = ../example.wsdl`
`include(../kdsoap.pri)`

`DEFINES += QT_DEPRECATED_WARNINGS`

`INCLUDEPATH += \`
       `$$KDSOAP_PATH/include \`
        `$$KDSOAP_PATH/include/KDSoapServer \`
       `$$KDSOAP_PATH/include/KDSoapClient`

`LIBS += -L$$KDSOAP_PATH/lib -lkdsoap -lkdsoap-server`
`#`  

And when I run qmake from Qt Creator I got:  kdsoap.pri:7: error: AND operator without prior condition.

Affected line 7 and 8 in kdsoap.pri contains:
`:  # copy from environment:`
`isEmpty( KDSOAPDIR ):KDSOAPDIR="$$(KDSOAPDIR)"`

I fixed it by comment out that colon like this...
`#:  # copy from environment:`
`isEmpty( KDSOAPDIR ):KDSOAPDIR="$$(KDSOAPDIR)"` 

But I am not sure if I corrupt something with this fix... Generator and compilation is working with this fix, everything seems to work.
Qt version 5.15.0, 5.14, Linux
Thank you

dfaure-kdab:
Thanks for the report, this stray ":" had no reason to be. Fixed.

However nowadays I would recommend using cmake, both for building KDSoap and for your own project :-)

PR #261, opened 2023-04-14
Remove stray ":" in .prf file

pocek:
Same story as in #229 and 08779c33df46579ae35859e223811ab318830da3.


dfaure-kdab (approved):
Thanks.

Same comment, too: why not cmake? ;-)

PR #273, opened 2024-01-03
Fix Qt5-Qt6 coinstall

### LINKED ISSUE #262: Co-installability of Qt5 and Qt6 version
Moin moin 

It would be great if KDSoap would be co-installable for both Qt5 and Qt6. 

It seems all files are suffixed appropriatly (or only present in one of the bindings) apart from the header files. Their co-installability was reverted in https://github.com/KDAB/KDSoap/commit/d2a04a2 .

Are there any plans to mitigate this issue, or how are consumers supposed to handle this as more and more, but not all, software moves to Qt6.

mfg Tobias

redstrate:
Fixes #262.

This moves Qt6 headers to KDSoapX-qt6, similar to how the rest of the Qt6 library is installed. This also fixes kdwsdl2cpp to point to the correct header (so qt6 will generate includes for -qt6 and Qt5 is still the plain header name.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,7 +207,11 @@ if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
         COMPATIBILITY
         AnyNewerVersion
     )
-    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/kdsoap_version.h" DESTINATION ${INSTALL_INCLUDE_DIR}/KDSoapClient)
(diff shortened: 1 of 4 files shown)


dfaure-kdab (changes requested):
Application code uses `#include <KDSoapClient/KDSoapClientInterface>` etc, please keep compatibility with that, e.g. using the approach I suggested in #262

dfaure-kdab (approved):

dfaure-kdab (approved):

PR #280, opened 2024-01-08
CI: Add CLAZY_CHECKS and CLAZY_IGNORE_DIRS env vars

dantti:


.github/workflows/nightly.yml
```diff
@@ -8,6 +8,22 @@ on:
...
+      - kdsoap-2.2
+  pull_request:
+    branches:
+      - master
+      - kdsoap-2.1
+      - kdsoap-2.2
+
+env:
+  CLAZY_CHECKS: "level2,no-qstring-allocations,no-fully-qualified-moc-types"
+  CLAZY_IGNORE_DIRS: ".libkode."
```
dfaure-kdab, line 25: It's a regexp, so it should be ".*libkode.*"
Or maybe it works with a substring, no idea, but at least .libkode. doesn't make sense to me.

.github/workflows/nightly.yml
```diff
@@ -8,6 +8,22 @@ on:
   schedule:
   - cron: '0 3 * * *'
 
+  # TODO remove
```
dfaure-kdab, line 11: ?

dantti: To force it to run now, I add this section, if it works as expected I removed before removing, otherwise we wait till 3am UTC :D

.github/workflows/nightly.yml
dantti: hmm I think github formatted what you wrote :smile:

winterz: \.\*libkode\.\*

.github/workflows/nightly.yml
```diff
@@ -8,6 +8,10 @@ on:
   schedule:
   - cron: '0 3 * * *'
 
+env:
+  CLAZY_CHECKS: "level2,no-qstring-allocations,no-fully-qualified-moc-types"
```
dfaure-kdab, line 12: https://invent.kde.org/sdk/clazy/-/merge_requests/111 just landed so we don't need to exclude fully-qualified-moc-types anymore -- assuming you can somehow make CI use the latest clazy from git master.

dantti: I use distro packages, if there were prebuilt clazy would likely be easy, we can also compile it here if you think it's worth it?

dfaure-kdab: Given the low amount of development on KDSoap, it's fine to exclude it until the distro has a fixed clazy.

PR #290, opened 2024-07-24
Require C++17, port from deprecated qAsConst to std::as_const

dfaure-kdab:


dantti:
Seems the issue is an old clang, some kdab project are using os-latest instead of specif OS versions, would that be ok?

dfaure-kdab:
No I think the issue is that std::random_shuffle was removed in C++17 so I need to port the code to something else.
https://en.cppreference.com/w/cpp/algorithm/random_shuffle

PR #292, opened 2024-08-21
Fix tests to adapt to Qt 6.8's QNAM sending lowercase headers

dfaure-kdab:


unittests/serverlib/test_serverlib.cpp
```diff
@@ -1343,10 +1343,18 @@ private Q_SLOTS:
         loop.exec();
         reply->readAll();
 
-        QVERIFY(reply->rawHeaderList().contains("Access-Control-Allow-Origin"));
-        QCOMPARE(reply->rawHeader("Access-Control-Allow-Origin").constData(), "*");
-        QVERIFY(reply->rawHeaderList().contains("Access-Control-Allow-Headers"));
-        QCOMPARE(reply->rawHeader("Access-Control-Allow-Headers").constData(), "Content-Type");
+        const bool titleCase = reply->rawHeaderList().contains("Access-Control-Allow-Origin");
+        const bool lowerCase = reply->rawHeaderList().contains("access-control-allow-origin");
```
dantti, line 1347: You should use hasRawHeader() which follows the spec on case-insensitive headers

dfaure-kdab: and then? How do I retrieve the value of the raw header?
Clearly reply->rawHeader("Access-Control-Allow-Origin") fails with newer Qt versions...

(I don't know if all-lowercase works with older Qt versions though)

dantti: well Qt5 source code does a case insensitive search so it should just work https://codebrowser.dev/qt5/qtbase/src/network/access/qnetworkrequest.cpp.html#_ZNK22QNetworkHeadersPrivate13findRawHeaderERK10QByteArray

dantti: Your initial issue was that `reply->rawHeaderList().contains("Access-Control-Allow-Origin")` operates with case-sesitivity, replacing with `reply->hasRawHeader("...");` should likely be the only fix here

dfaure-kdab: Yes but how do I then port `QCOMPARE(reply->rawHeader("Access-Control-Allow-Headers").constData(), "Content-Type");` ?

dantti: so `reply->rawHeader("Access-Control-Allow-Headers")` should give the content of "Access-Control-Allow-Headers" or "access-ConTROL-aLLow-headers" as it calls that findRawHeader function which is case-insensitive.
I don't know why that would fail, if the issue is with Qt then it's a Qt bug.

dfaure-kdab: Oh I see. I misunderstood how this works. You're right. I'll submit a followup.

PR #305, opened 2025-02-26
Add CMakePresets.json

iamsergio:
Decided not to clutter the file with Qt 5 variants. Whoever needs can add them, but IMHO the Qt5 CI coverage is enough to guarantee Qt5 works.


dfaure-kdab:
Good thing I forgot to `git add` mine, yours is much better ;)

iamsergio:
Will copy and adapt to KDReports

PR #307, opened 2025-03-01
CI: Enable -Werror for clazy since we're warning clean now

iamsergio:


CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ option(${PROJECT_NAME}_TESTS "Build the tests" OFF)
 option(${PROJECT_NAME}_EXAMPLES "Build the examples" ON)
 option(${PROJECT_NAME}_DOCS "Build the API documentation" OFF)
 option(${PROJECT_NAME}_QT6 "Build against Qt 6" ON)
+option(${PROJECT_NAME}_WERROR "Treat build warnings as errors" OFF)
```
dfaure-kdab, line 77: CMake provides a builtin option for that already: CMAKE_COMPILE_WARNING_AS_ERROR

I learned that very recently in https://codereview.kdab.com/c/training-material/+/151437

Should we just document to use that instead?

iamsergio: good idea

iamsergio: the preset serves as documentation

CMakeLists.txt
```diff
@@ -102,6 +102,20 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     )
 endif()
 
+if(CMAKE_COMPILE_WARNING_AS_ERROR)
+    # These -Wno-error flags are here to enable -Werror for clazy in CI
+    # it doesn't mean we endorse those warnings.
+    # MSVC Werror is untested and probably will have a few hundered errors
+    if(MSVC)
+        add_compile_options(/WX)
```
dfaure-kdab, line 110: This can be removed the cmake option takes care of that

Modules/Compiler/MSVC.cmake:13:   set(CMAKE_${lang}_COMPILE_OPTIONS_WARNING_AS_ERROR "-WX")

PR #309, opened 2025-03-17
Add cppcheck target and call it on CI

dantti:
It requires
CMAKE_EXPORT_COMPILE_COMMANDS=ON
which is enabled on base preset.

### DIFF
--- a/.github/workflows/tidyclazy.yml
+++ b/.github/workflows/tidyclazy.yml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MIT
 
-name: Tidy and Clazy
+name: Tidy / Clazy / cppcheck
 
 on:
   push:
@@ -35,6 +35,7 @@ jobs:
             qt_version: "6.6"
             apt_pgks:
               - clazy
+              - cppcheck
 
     steps:
       - name: Install Qt ${{ matrix.config.qt_version }} with options and default aqtversion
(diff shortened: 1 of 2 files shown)


dantti:
@dfaure @iamsergio clazy is not WError, but cppcheck is should I make it just report the issues? I think with this we can retire our CI builders for it

iamsergio:
> @dfaure @iamsergio clazy is not WError, but cppcheck is should I make it just report the issues? I think with this we can retire our CI builders for it

Clazy is Werror already actually.

cppcheck should just report the errors (so CI passes), until someone curates the cppcheck warning list.

you can pass `-DCMAKE_COMPILE_WARNING_AS_ERROR=OFF --target cppcheck`

dantti:
then CMAKE_COMPILE_WARNING_AS_ERROR is not working as I said, you can see it reached cppcheck while there were several warnings in clazy first.

iamsergio:
> then CMAKE_COMPILE_WARNING_AS_ERROR is not working as I said, you can see it reached cppcheck while there were several warnings in clazy first.

Can you double-check ? I don't see the clazy warnings, note that the `-Wdocumentation` spam doesn't come from clazy and we have a `Wno-error=documentation` for that

dantti:
ok, deprecated-declarations seemed valid but maybe too much work that might be better done on Qt5 support is dropped.
Then this PR should be fine now, and I can clean our CI

iamsergio (approved):
```
CMakeLists.txt:        add_compile_options(-Wno-error=deprecated-declarations)
CMakeLists.txt:            add_compile_options(-Wno-error=documentation -Wno-error=unused-private-field -Wno-error=c++20-extensions)

```
yeah deprecated-declarations also ignored

PR #315, opened 2025-06-17
Improve security when handling requests to download files

### LINKED ISSUE #314: Path is not cleaned fully.
This is similar to #247 I just never noticed it then because I was using a web browser to test and it sanitizes what it sends.

I initially tested this using curl -s --path-as-is http://127.0.0.1/../../../home/user/.vnc/vncpasswd

Some more unittests for testFileDownload_data() in the serverlib unittest:

```
        QTest::newRow("leading_double_slash") << "///path/to/file_download.txt" << readable << httpOK;
        QTest::newRow("leading_triple_slash") << "///path/to/file_download.txt" << readable << httpOK;
        QTest::newRow("leading_triple_slash_and_dot_dot") << "///../path/to/file_download.txt" << readable << httpOK;
        QTest::newRow("leading_double_slash_and_dot_dot") << "//../path/to/file_download.txt" << readable << httpOK;
        QTest::newRow("leading_slash_and_dot_dot") << "/../path/to/file_download.txt" << readable << httpOK;
```

The above paths are all passed to the processFileRequest function unmodified. QDir::cleanPath will leave upto three leading slashes and as many .. at the start of the URL as there are.

My fix for this is:

    QString path = QDir::cleanPath(url.path());
    path = path.mid(path.indexOf(QRegularExpression(R"((?:\/$|\/[^\/.]+))")));

I don't know whether you want to use QRegularExpression or not though. Basically that boils down to strip the start of the string and accept either / on it's own or any slash that isn't followed by a . or another /.
That does fix the above, but the likes of /.wsdl/ would be removed.

dfaure-kdab:
Make sure we never go outside the base directory

Fixes #314

### DIFF
--- a/docs/CHANGES_2_3.txt
+++ b/docs/CHANGES_2_3.txt
@@ -12,3 +12,4 @@ Server-side:
 ============
 * To avoid mixing up raw-xml requests in the same server object (#288), KDSoap now creates a server object for each incoming connection.
   Make sure your server object is ready to be created multiple times (this was already a requirement when enabling multi-threading with setThreadPool()).
+* Improve security when handling requests to download files, to make sure we never go outside the base directory (#314)
--- a/src/KDSoapServer/KDSoapServerSocket.cpp
+++ b/src/KDSoapServer/KDSoapServerSocket.cpp
@@ -83,8 +83,12 @@ static HeadersMap parseHeaders(const QByteArray &headerData)
     const QByteArray path = queryPos >= 0 ? arg1.left(queryPos) : arg1;
     const QByteArray query = queryPos >= 0 ? arg1.mid(queryPos) : QByteArray();
     // Unfortunately QDir::cleanPath works with QString
-    const QByteArray cleanedPath = QDir::cleanPath(QString::fromUtf8(path)).toUtf8();
-    headersMap.insert("_path", cleanedPath + query);
+    QByteArray cleanedPath = QDir::cleanPath(QString::fromUtf8(path)).toUtf8();
+    // And unfortunately it keeps "//<host>/" unchanged on Windows
+    while (cleanedPath.startsWith("//")) // while() so we also clean up "///"
+        cleanedPath = cleanedPath.mid(1);
+    headersMap.insert("_path", cleanedPath);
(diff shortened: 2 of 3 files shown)


jtjbrady (commented):
In case you're wondering why the unit test is failing on Windows but passing on Linux.

The first two unit tests start with /// this is unmodified by QDir::cleanPath on Windows (due to UNC paths I believe), on Linux this becomes a single /

So that the server performs the same on Linux and Windows I believe multiple leading slashes should be stripped in parseHeaders.

dfaure-kdab:
> In case you're wondering why the unit test is failing on Windows but passing on Linux.
> 
> The first two unit tests start with /// this is unmodified by QDir::cleanPath on Windows (due to UNC paths I believe), on Linux this becomes a single /
> 
> So that the server performs the same on Linux and Windows I believe multiple leading slashes should be stripped in parseHeaders.

Thanks, I've been failing to get the output from the failing test on Windows (and I'm on Linux myself).
However I don't really understand: given the `Qt::SkipEmptyParts` argument to split, a path with "///" as the result of cleanPath should still have no impact on the rest of the code. In fact the test passes on Linux even if I remove the call to `QDir::cleanPath`.

jtjbrady:
qdir.cpp line 66 in rootLength has Windows specific rules to do with UNC paths.  Those are normally //servername/
Apparently a blank server name is allowed which gives us three slashes.

dfaure-kdab:
Oh, I missed the fact that `qt_normalizePathSegments` called `rootLength`.
Thanks for the analysis. All clear now, here's an updated MR.

dfaure-kdab:
OK, it passed. What do you think of this solution?

dfaure-kdab:
Good point. But we've already split out the query in parseHeaders, and then reassembled it, I think I'll rather store them split and reassemble afterwards in handleRequest. No point in parsing twice (with risks of slight differences).

Also, since consecutive slashes have been cleaned up, we can just do `path.startsWith(QLatin1String("/.."))`

jtjbrady:
Only remaining issue I see is the one in the code review above.

The unit test "leading_double_slash" is identical to "leading_triple_slash".

PR #318, opened 2025-06-23
Support for RFC 7233 (Range Requests).

jtjbrady:
This pull request adds support for HTTP range requests to handleFileDownload.

Range requests allow a client to request only a portion of a file to be transferred. What I actually use this for is for transferring data or log files which grow over time, rather than retreiving the whole file every time, this allows the client to resume the download.

If you're happy to add support for range requests I'll write some unit tests.

### DIFF
--- a/src/KDSoapServer/KDSoapServerSocket.cpp
+++ b/src/KDSoapServer/KDSoapServerSocket.cpp
@@ -26,7 +26,6 @@
 #include <QThread>
 #include <QUuid>
 #include <QVarLengthArray>
-#include <cmath>
 
 static const char s_forbidden[] = "HTTP/1.1 403 Forbidden\r\nContent-Length: 0\r\n\r\n";
 
@@ -440,18 +439,18 @@ bool KDSoapServerSocket::handleWsdlDownload(KDSoapServerObjectInterface *serverO
     return false;
 }
 
-std::optional<QVector<QPair<int, int>>> KDSoapServerSocket::determineFileRanges(int fileSize)
+KDSoapServerSocket::FileRange KDSoapServerSocket::determineFileRanges(int fileSize) const
 {
     QVector<QPair<int, int>> requestedRanges;
 
     const auto headerIt = m_httpHeaders.constFind("range");
     if (headerIt == m_httpHeaders.constEnd())
-        return requestedRanges; // No Range header present — full file response
+        return {FileRange::Type::FullFile}; // No Range header present — full file response
 
     const QByteArray rangeHeader = headerIt.value().trimmed();
     if (!rangeHeader.startsWith("bytes=")) {
         // Section 3.1 (RFC 7233): Ignore a range header field that contains a range unit the server does not understand
-        return requestedRanges;
+        return {FileRange::Type::FullFile};
     }
 
     const QByteArray rangeSpec = rangeHeader.mid(6);
@@ -461,7 +460,7 @@ std::optional<QVector<QPair<int, int>>> KDSoapServerSocket::determineFileRanges(
         // Section 14.3.2: A server that receives a malformed Range header field
         // MAY ignore it and serve the whole resource, or respond with a 416
         // We ignore it and send a 200.
-        return requestedRanges;
+        return {FileRange::Type::FullFile};
     }
 
     const QByteArrayList rangeSpecs = rangeHeader.mid(6).split(',');
@@ -510,12 +509,26 @@ std::optional<QVector<QPair<int, int>>> KDSoapServerSocket::determineFileRanges(
 
     // Sort and coalesce ranges
     if (requestedRanges.size() > 1) {
+        // Section 3.1 (RFC 7233): Says the client SHOULD send the ranges in order,
+        // unless it has a valid reason not to.
+        // Section 4.1 (RFC 7233): Says the server SHOULD send the parts in the
+        // same order the client requested. It also states that the client MUST
+        // inspect the Content-Range header and cannot rely on receiving the ranges
+        // in the same order as requested.
+
+        // Section 4.1 (RFC 7233): "When multiple ranges are requested, a server MAY coalesce
+        // any of the ranges that overlap, or that are separated by a gap that is smaller
+        // than the overhead of sending multiple parts, regardless of the order in which
+        // the corresponding byte-range-spec appeared in the received Range header field."
+        constexpr int multiPartOverhead = 86; // Minimum realistic overhead
+
+        // We sort so that we can coalesce, even though that changes the order of the ranges.
         std::sort(requestedRanges.begin(), requestedRanges.end());
 
         for (int i = 0; i < requestedRanges.size() - 1;) {
             auto &current = requestedRanges[i];
             auto &next = requestedRanges[i + 1];
-            if (next.first <= current.second + 1) {
+            if (next.first <= current.second + 1 + multiPartOverhead) {
                 current.second = qMax(current.second, next.second);
                 requestedRanges.remove(i + 1);
             } else {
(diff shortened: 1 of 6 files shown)


jtjbrady:
Yes, I spotted that one, I guess you don't see the "review", maybe it only goes to the author, so I asked myself to review it?
What I said was:

Is the correct change to stop clazy complaining about this

`next = requestedRanges.erase(QVector<QPair<int, int>>::const_iterator(next));`

I'm assuming this is a clazy false positive or am I missing something?

That part of the code is:

```
                    // Coalesce
                    for (auto first = requestedRanges.begin(), next = first + 1; next != requestedRanges.end();) {
                        if (next->first <= first->second) {
                            // Next is directly after (or overlaps) first, so merge it into first.
                            first->second = qMax(first->second, next->second);
                            next = requestedRanges.erase(next);
                        } else {
                            first = next;
                            ++next;
                        }
                    }
```

As far as I am concerned "first" has to be an iterator not a const_iterator because of the line first->second = qMax(first->second, next->second);
The erase function has always confused me to a degree by taking a const_iterator parameter and returning an iterator.

dfaure-kdab:
> The erase function has always confused me to a degree by taking a const_iterator parameter and returning an iterator.

Then clazy is technically correct, this code does convert a const_iterator into an iterator ;-) But yes it's a kind of false positive because it's a case where it doesn't matter (IIRC the goal of detecting these conversions is to avoid implicit detaching in iterator-based for loops).

We have the same issue in our in-house application, you can add the same suppression:

    it = m_entries.erase(it); // clazy:exclude=strict-iterators

src/KDSoapServer/KDSoapServerSocket.cpp
```diff
@@ -414,22 +424,126 @@ void KDSoapServerSocket::handleRequest(const QMap<QByteArray, QByteArray> &httpH
...
+    }
+
+    if (requestedRanges.isEmpty() || rangeSpecs.size() > 5) {
+        // No ranges found in header or
+        // Section 6.1 (RFC 7233): Too many ranges — reject
+        write("HTTP/1.1 416 Range Not Satisfiable\r\nContent-Length: 0\r\n\r\n");
+        return std::nullopt;
+    }
+
+#if 0
```
dfaure-kdab, line 534: Should this code be in or out? I.e. if it's not wanted, why not delete it?

src/KDSoapServer/KDSoapServerSocket.cpp
```diff
@@ -444,7 +558,123 @@ bool KDSoapServerSocket::handleFileDownload(KDSoapServerObjectInterface *serverO
...
+        response.reserve(50);
+        response += "HTTP/1.1 206 Partial Content\r\n";
+        response += "Accept-Ranges: bytes\r\n";
+
+        auto writeRange = [this, device](const QPair<int, int> range) {
+            device->seek(range.first);
+
+            char block[4096] = {0};
+            qint64 totalRead = 0, toRead = range.second - range.first + 1;
+            ;
```
dfaure-kdab, line 576: stray semicolon

src/KDSoapServer/KDSoapServerSocket.cpp
```diff
@@ -444,7 +558,123 @@ bool KDSoapServerSocket::handleFileDownload(KDSoapServerObjectInterface *serverO
         delete device;
         return true; // handled!
     }
-    const QByteArray response = httpResponseHeaders(false, contentType, device->size(), m_serverObject);
+
+    if (auto r = determineFileRanges(device->size()); !r) {
+        delete device;
+        return true;
+    } else if (const QVector<QPair<int, int>> requestedRanges = *r; !requestedRanges.isEmpty()) {
```
dfaure-kdab, line 565: This method is getting very long, could the body of this if() be moved to a separate method?

src/KDSoapServer/KDSoapServerSocket.cpp
```diff
@@ -414,22 +424,126 @@ void KDSoapServerSocket::handleRequest(const QMap<QByteArray, QByteArray> &httpH
...
+        // MAY ignore it and serve the whole resource, or respond with a 416
+        // We ignore it and send a 200.
+        return requestedRanges;
+    }
+
+    const QByteArrayList rangeSpecs = rangeHeader.mid(6).split(',');
+    for (const QByteArray &spec : rangeSpecs) {
+        const QList<QByteArray> parts = spec.trimmed().split('-');
+        if (parts.size() != 2) {
+            requestedRanges.clear();
```
dfaure-kdab, line 471: This code is rather confusing to read because for the caller of this method, an empty requestedRanges (as returned on lines 449, 454 and 464) means "full file response".
Starting from this point however, an empty requestedRanges means error 416 (which gets returned as a nullopt for the caller). So the same thing (empty vector) means something else (temporarily), it's a bit confusing and error-prone in my opinion.

Even from the caller's point of view I was already finding empty-vector=full-file to be unclear.
I'm not sure what the perfect solution would be, maybe returning a struct { FileRangeType type; RangeVector ranges; }` with `enum FileRangeType { InvalidRange, FullFile, ValidRanges }` and `using RangeVector = QVector<QPair<int, int>>` ?
In case of ValidRanges, we'd assert that the vector isn't empty.

src/KDSoapServer/KDSoapServerSocket.cpp
```diff
@@ -414,22 +424,126 @@ void KDSoapServerSocket::handleRequest(const QMap<QByteArray, QByteArray> &httpH
...
+            } else {
+                ++i;
+            }
+        }
+    }
+
+    if (requestedRanges.isEmpty() || rangeSpecs.size() > 5) {
+        // No ranges found in header or
+        // Section 6.1 (RFC 7233): Too many ranges — reject
+        write("HTTP/1.1 416 Range Not Satisfiable\r\nContent-Length: 0\r\n\r\n");
```
dfaure-kdab, line 530: shouldn't the caller take care of this write()? I would expect this method to only do the parsing (and to be `const`).

src/KDSoapServer/KDSoapServerSocket.cpp
```diff
@@ -444,7 +558,123 @@ bool KDSoapServerSocket::handleFileDownload(KDSoapServerObjectInterface *serverO
...
         return true; // handled!
     }
-    const QByteArray response = httpResponseHeaders(false, contentType, device->size(), m_serverObject);
+
+    if (auto r = determineFileRanges(device->size()); !r) {
+        delete device;
+        return true;
+    } else if (const QVector<QPair<int, int>> requestedRanges = *r; !requestedRanges.isEmpty()) {
+        QByteArray response;
+        response.reserve(50);
```
dfaure-kdab, line 567: Good idea, but surely not enough, looks like it could be 500 ;)

unittests/ranges/test_ranges.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+    void testPartiallyUnsatisfiableRanges();
+
+    void testUnsatisfiableRanges_data();
+    void testUnsatisfiableRanges();
+
+    void testRangeWithWhitespace_data();
+    void testRangeWithWhitespace();
+
+#if 0
```
dfaure-kdab, line 95: same question here. Unfinished work?

unittests/ranges/test_ranges.cpp
```diff
@@ -0,0 +1,497 @@
...
+        }
+    } else if (status == 200) {
+        // For 200 OK full content, expect exact data match
+        QCOMPARE(reply->readAll(), testFileData);
+    } else {
+        // For 416 or others, body should be empty
+        QCOMPARE(reply->readAll().size(), 0);
+    }
+
+    reply->deleteLater();
```
dfaure-kdab, line 203: I think this could be `delete reply`, we're not in a slot connected to reply, and there's not really an event loop running to handle that deleteLater (they still all get deleted before exit, I think)

(repeats)

unittests/ranges/test_ranges.cpp
```diff
@@ -0,0 +1,497 @@
...
+            QCOMPARE(status, 206);
+        } else
+            QCOMPARE(status, expectedStatusCode);
+        const QByteArray contentTypeHeader = reply->header(QNetworkRequest::ContentTypeHeader).toByteArray();
+        QVERIFY(!contentTypeHeader.startsWith("multipart/byteranges"));
+        const QByteArray data = reply->readAll();
+        QCOMPARE(data, expectedRanges.first());
+    } else {
+        QCOMPARE(status, 206);
+        QCOMPARE(status, expectedStatusCode);
```
dfaure-kdab, line 265: How does this relate to the previous line?

Are we expecting 206 or expectedStatusCode, or are both always equal?

This feels a bit like a unittest for the unittest ;-)

(same on lines 251/252).

If they're always equal, I'd just do `QCOMPARE(status, expectedStatusCode);` globally (i.e. I'd trust the `_test()` method to have the proper expected values), with the 200-or-206 caveat (maybe the test() method could provide a list instead of a single value...).

dfaure-kdab (changes requested):
This is turning into a full-fledged HTTP server ;-)

unittests/ranges/test_ranges.cpp
jtjbrady: More of a lack of support for HEAD requests in KDSoap - yes KDSoapCustomVerbRequestInterface is available, but it doesn't fit in very well with file requests.  The tests are however complete.

src/KDSoapServer/KDSoapServerSocket.cpp
jtjbrady: Agreed, I much prefer this new implementation after your suggestion.

unittests/ranges/test_ranges.cpp
dfaure-kdab: Ah. Can you add a comment next to the `#if 0`, so the next reader understands why this is commented out?

The alternative is to use QEXPECT_FAIL so that if/when someone fixes things, the tests will automatically show up as suddenly passing unexpectedly (and so that we don't regress e.g. to the point of crashing when receiving a HEAD request).

